PR #3, opened 2016-03-01
Initial ramblings about ConfigValidator

leonerd:


leonerd:
(oops; slip of the mouse)

HOWTO.md
```diff
@@ -287,9 +287,41 @@ new Cli({
...
 ```
 
+# Configuration
+So far in this example we have hard-coded various items of information that would be
+considered "configuration"; namely the Slack outbound webhook token and the list of room
+mappings to bridge. We can use the `ConfigValidator` to help parse a configuration file
+at startup time to obtain this information from instead.
+
+Start by defining a schema file that describes what the YAML config file can contain.
+This is also a YAML file. Store this in a file called `slack-config-schema.yaml`:
```
Kegsay: Should mention that this follows JSON Schema.

HOWTO.md
```diff
@@ -287,9 +287,41 @@ new Cli({
...
+```yaml
+type: object
+requires: ["slack_webhook_url"]
+properties:
+    slack_webhook_url:
+        type: string
+```
+
+If we supply the name of this schema file to the constructor of the main `Cli` object
+then it will use this to valid a config file that the user passes on the commandline.
```
Kegsay: s/valid/validate/

Kegsay: Also s/commandline/command line/

Kegsay:
LGTM aside from nits.

PR #11, opened 2016-07-08
Add `StateLookup` component

Kegsay:
This is designed in part to fix https://github.com/matrix-org/matrix-appservice-irc/issues/41 - but is a generally useful component for Bridges to have access to.


Kegsay:
ok to test

PR #13, opened 2016-08-25
Better scheduling of events sent to HS

lukebarnard1:
Previously, the bridge bot may become head-of-line blocked by it's own event scheduling back-off. This has been made less aggressive; the delay between each request to send a previously failed sending of event increases linearly now.

Also, the queuing algorithm separates queues by room ID.

Fix for https://github.com/matrix-org/matrix-appservice-irc/issues/122

### DIFF
--- a/lib/components/client-factory.js
+++ b/lib/components/client-factory.js
@@ -13,6 +13,9 @@ var requestModule = require("request");
  * @param {string=} opts.appServiceUserId The application service's user ID. Must
  * be set prior to calling getClientAs(). See configure() to set this after
  * instantiation.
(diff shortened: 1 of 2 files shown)


Kegsay:
ok to test

lib/components/client-factory.js
```diff
@@ -22,6 +22,7 @@ function ClientFactory(opts) {
     //      user_id: Client
     //  }
     };
+    this._clientScheduler = opts.clientScheduler;
```
Kegsay: Docs!

Kegsay:
LGTM :star2:

PR #14, opened 2016-08-30
DB upgrade script

leonerd:
* Added a script to help AS authors upgrade databases from 0.3 to 1.0
 * Update documentation in `room-bridge-store.js` to more accurately reflect the data shape it creates

### DIFF
--- a/scripts/upgrade-db-0.3-to-1.0.js
+++ b/scripts/upgrade-db-0.3-to-1.0.js
@@ -43,14 +43,14 @@ process.exit(0);
 // ** You will probably want to customise these functions **
 // *********************************************************
 
-function generateMatrixID(opts)
+function generateMatrixEntry(opts)
 {
-    return opts.matrix_id;
+    return null;
 }
 
-function generateRemoteID(opts)
+function generateRemoteEntry(opts)
 {
-    return opts.remote_id;
+    return null;
 }
 
 function generateLinkId(opts)
@@ -106,11 +106,9 @@ var upgradeRooms = Promise.coroutine(function*(db) {
                 }
                 matrixRooms[e.id] = e.data;
 
-                var id = generateMatrixID({matrix_id: e.id, matrix: e.data});
-                insert(id, {
-                    matrix_id: e.id,
-                    matrix: e.data,
-                });
+                var entry = generateMatrixEntry({matrix_id: e.id, matrix: e.data});
+                if (entry) insert(entry.id, entry);
+
                 break;
             case "remote":
                 if (remoteRooms[e.id]) {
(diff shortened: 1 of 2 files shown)


Kegsay:
ok to test

scripts/upgrade-db-0.3-to-1.0.js
```diff
@@ -0,0 +1,199 @@
...
+   --rooms   The path to rooms.db. Required.`
+);
+process.exit(0);
+}
+
+// *********************************************************
+// ** You will probably want to customise these functions **
+// *********************************************************
+
+function generateMatrixEntry(opts)
```
Kegsay: Please doc what properties exist in `opts`.

If they are all the same then you can do it once in `:43`, but as of right now I got no clue.

scripts/upgrade-db-0.3-to-1.0.js
```diff
@@ -0,0 +1,199 @@
...
+}
+
+function generateRemoteEntry(opts)
+{
+    return null;
+}
+
+function generateLinkId(opts)
+{
+    return opts.matrix_id + " " + opts.remote_id;
```
Kegsay: A one-liner explaining that the delimiter must not be a valid character in the `remote_id` format would be good.

scripts/upgrade-db-0.3-to-1.0.js
```diff
@@ -0,0 +1,199 @@
...
+if (!opts.help && !opts.rooms) {
+    console.log("--rooms is required.");
+    opts.help = true;
+}
+
+if (opts.help) {
+    console.log(
+`Database Upgrade script (v0.3 => v1.0)
+--------------------------------------
+
```
Kegsay: Where does this dump the result to?

scripts/upgrade-db-0.3-to-1.0.js
```diff
@@ -0,0 +1,199 @@
...
+            sparse: true
+        });
+    } catch (err) {
+        console.error(JSON.stringify(err));
+    }
+});
+
+Promise.coroutine(function*() {
+    try {
+        fs.mkdirSync("1.0-db");
```
Kegsay: `rooms-1.0.db` may be a better name (`.db` file extension).

Kegsay: Oh ignore me, this is the directory

Kegsay:
LGTM aside from a tiny bit of extra documentation (given this is supposed to be read as a guide for upgrading).

scripts/upgrade-db-0.3-to-1.0.js
leonerd: Done 89ce7c8

scripts/upgrade-db-0.3-to-1.0.js
leonerd: Doced d819ec4

PR #15, opened 2016-08-31
Provide wrapper for 3PE lookup APIs

leonerd:
This PR adds a new `controller` feature, `onThirdPartyLookup` which helps bridges to implement the API defined in
  https://docs.google.com/document/d/13NGa46a_WWno-XYfe8mQrglQdtOYMFVZtxkPKXDC3ac/edit#heading=h.m0btedxhv6ao

### DIFF
--- a/lib/bridge.js
+++ b/lib/bridge.js
@@ -268,7 +268,7 @@ Bridge.prototype.customiseAppservice = function(appService) {
             });
         }
 
-        app.get("/_matrix/app/:ver/thirdparty/protocol/:protocol", (req, res) => {
+        app.get("/_matrix/app/:ver/thirdparty/protocol/:protocol", function(req, res) {
             if (req.params.ver != "unstable") {
                 res.status(404).send("Cannot GET " + req.path + "\r\n");
                 return;
@@ -287,12 +287,12 @@ Bridge.prototype.customiseAppservice = function(appService) {
             }
 
             lookupController.getProtocol(protocol).then(
-                (result) => { res.status(200).json(result) },
-                (e)      => { res.status(e.code).json({error: e.err}) }
+                function(result) { res.status(200).json(result) },
+                function(e) { res.status(e.code).json({error: e.err}) }
             );
         });
 
-        app.get("/_matrix/app/:ver/thirdparty/location/:protocol", (req, res) => {
+        app.get("/_matrix/app/:ver/thirdparty/location/:protocol", function(req, res) {
             if (req.params.ver != "unstable") {
                 res.status(404).send("Cannot GET " + req.path + "\r\n");
                 return;
@@ -311,12 +311,12 @@ Bridge.prototype.customiseAppservice = function(appService) {
             }
 
             lookupController.getLocation(protocol, req.query).then(
-                (result) => { res.status(200).json([result]) },
-                (e)      => { res.status(e.code).json({error: e.err}) }
+                function(result) { res.status(200).json(result) },
+                function(e) { res.status(e.code).json({error: e.err}) }
             );
         });
 
-        app.get("/_matrix/app/:ver/thirdparty/user/:protocol", (req, res) => {
+        app.get("/_matrix/app/:ver/thirdparty/user/:protocol", function(req, res) {
             if (req.params.ver != "unstable") {
                 res.status(404).send("Cannot GET " + req.path + "\r\n");
                 return;
@@ -335,8 +335,8 @@ Bridge.prototype.customiseAppservice = function(appService) {
             }
 
             lookupController.getUser(protocol, req.query).then(
-                (result) => { res.status(200).json([result]) },
-                (e)      => { res.status(e.code).json({error: e.err}) }
+                function(result) { res.status(200).json(result) },
+                function(e) { res.status(e.code).json({error: e.err}) }
             );
         });
     }
--- a/lib/bridge.js
+++ b/lib/bridge.js
@@ -857,8 +857,8 @@ BridgeContext.prototype.get = function(roomStore, userStore) {
   * Invoked on requests for 3PE query metadata
   * @callback Bridge~getProtocol
   * @param {string} protocol The name of the 3PE protocol to query
-  * @return {Bridge~thirdPartyProtocolResult} Metadata about 3PE queries that
-  * can be made for this protocol.
+  * @return {Promise<Bridge~thirdPartyProtocolResult>} A Promise of metadata
+  * about 3PE queries that can be made for this protocol.
   */
 
  /**
@@ -876,7 +876,8 @@ BridgeContext.prototype.get = function(roomStore, userStore) {
   * @param {string} protocol The name of the 3PE protocol
   * @param {Object} fields The location query field data as specified by the
   * specific protocol.
-  * @return {Bridge~thirdPartyLocationResult} A 3PL lookup result
+  * @return {Promise<Bridge~thirdPartyLocationResult>} A Promise of a 3PL
+  * lookup result.
   */
 
  /**
@@ -895,7 +896,8 @@ BridgeContext.prototype.get = function(roomStore, userStore) {
   * @param {string} protocol The name of the 3PE protocol
   * @param {Object} fields The user query field data as specified by the
   * specific protocol.
-  * @return {Bridge~thirdPartyUserResult} A 3PU lookup result
+  * @return {Promise<Bridge~thirdPartyUserResult>} A Promise of a 3PU lookup
+  * result.
   */
 
  /**
--- a/lib/bridge.js
+++ b/lib/bridge.js
@@ -262,61 +262,73 @@ Bridge.prototype._customiseAppservice = function() {
             }
         }
 
-        this.addAppServicePath("GET", "/_matrix/app/:ver/thirdparty/protocol/:protocol", function(req, res) {
-            if (req.params.ver !== "unstable" || !lookupController.getProtocol) {
-                res.status(404).send("Cannot GET " + req.path + "\r\n");
-                return;
-            }
+        this.addAppServicePath({
+            method: "GET",
+            path: "/_matrix/app/:ver/thirdparty/protocol/:protocol",
+            handler: function(req, res) {
+                if (req.params.ver !== "unstable" || !lookupController.getProtocol) {
+                    res.status(404).send("Cannot GET " + req.path + "\r\n");
+                    return;
+                }
 
-            var protocol = req.params.protocol;
+                var protocol = req.params.protocol;
 
-            if (protocols.length && protocols.indexOf(protocol) === -1) {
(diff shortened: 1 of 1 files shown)


Kegsay:
ok to test

lib/bridge.js
```diff
@@ -742,3 +843,36 @@ BridgeContext.prototype.get = function(roomStore, userStore) {
...
+  * @callback Bridge~getLocation
+  * @param {string} protocol The name of the 3PE protocol
+  * @param {Object} fields The location query field data
+  */
+
+ /**
+  * Invoked on requests for 3PUs
+  * @callback Bridge~getUser
+  * @param {string} protocol The name of the 3PE protocol
+  * @param {Object} fields The user query field data
```
Kegsay: I assume the properties in `fields` are documented in the spec or something? Can you confirm?

lib/bridge.js
```diff
@@ -742,3 +843,36 @@ BridgeContext.prototype.get = function(roomStore, userStore) {
...
+  * @callback Bridge~getProtocol
+  * @param {string} protocol The name of the 3PE protocol to query
+  */
+
+ /**
+  * Invoked on requests for 3PLs
+  * @callback Bridge~getLocation
+  * @param {string} protocol The name of the 3PE protocol
+  * @param {Object} fields The location query field data
+  */
```
Kegsay: Do these really have no `@return`? If they don't, then please do not call them `getXYZ` as that implies something is being returned.

lib/bridge.js
```diff
@@ -237,11 +240,109 @@ Bridge.prototype.run = function(port, config, appServiceInstance) {
...
+    this.customiseAppservice(this.appService);
     this.appService.listen(port);
     return this.loadDatabases();
 };
 
 /**
+ * Apply any customisations required on the appService object.
+ * Subclasses are intended to override this method where appropriate. If
+ * overridden the subclass method will still have to call the one in this
+ * class.
```
Kegsay: Whilst the `Bridge` class *supports* subclassing, I dislike *advocating* it. There's a number of reasons for this:
 - You need to remember to call the parent method, as you mention here. This is easily missed.
 - It's not clear what can be subclassed and what cannot, nor is it clear *in the subclass* what is an overridden method. There is no visibility keyword or `@Override` annotations.
 - You ultimately need to know the inner workings of the parent method in order to fully utilise child methods. It's not clear what happens in `super.customiseAppservice` at all without actually digging through the code.
 - Javascript in particular has no real concept of inheritance, just prototypes, so trying to emulate inheritance feels a bit weird (though to be fair, ES6 are actively attempting to emulate it :disappointed: ).

From looking at the code, it looks like your **intention** is to allow `Bridge` subclasses the ability to add new paths to the `appService` itself. Why not expose that functionality instead e.g:

```javascript
// handlerFn = function(req, res){}
Bridge.prototype.addAppServicePath = function(method, path, handlerFn) {
    this.appService.app[method.toLowerCase()](path, handlerFn);
};
```

Express is [a little bit gross](http://expressjs.com/en/api.html#app.METHOD) with how you select the methods, but \*shrug\*. This makes it much clearer what "customisation" is being done, and keeps the library in control of *how* this handler function gets added to the `AppService` object in case we want to change it later down the line.

lib/bridge.js
```diff
@@ -237,11 +240,109 @@ Bridge.prototype.run = function(port, config, appServiceInstance) {
...
+Bridge.prototype.customiseAppservice = function(appService) {
+    // TODO(paul): This is gut-wrenching into the AppService instance itself.
+    //   Maybe an API on that object would be good?
+    var app = appService.app;
+
+    if (this.opts.controller.thirdPartyLookup) {
+        var lookupController = this.opts.controller.thirdPartyLookup;
+
+        var protocolFilter;
+        if ("protocols" in lookupController) {
```
Kegsay: `if (lookupController.protocols) {` is a better way of saying this because it is a static property name.

It looks like you're setting up a map with which to filter protocols. Given how few protocols we expect to be present (1 typically, maybe 2?), I would just linear scan and keep it as it is (a string array). This means you can remove `protocolFilter` and just replace checks with:

```javascript
if (lookupController.protocols.indexOf($PROTOCOL) === -1) {
    // not present
}
```
And do `lookupController.protocols.length === 0` to see if it was specified in the first place. You will need to default initialise it though in order to make sure that `protocols` is defined:
```javascript
lookupController.protocols = lookupController.protocols || [];
```

lib/bridge.js
```diff
@@ -237,11 +240,109 @@ Bridge.prototype.run = function(port, config, appServiceInstance) {
...
+            protocolFilter = {};
+            lookupController.protocols.forEach(function (p) {
+                protocolFilter[p] = true;
+            });
+        }
+
+        app.get("/_matrix/app/:ver/thirdparty/protocol/:protocol", (req, res) => {
```
Kegsay: The bridge library does not use ES6, so no arrow functions please!

lib/bridge.js
```diff
@@ -237,11 +240,109 @@ Bridge.prototype.run = function(port, config, appServiceInstance) {
...
+            if (req.params.ver != "unstable") {
```
Kegsay: Nit: Default to `!==` and only use `!=` if you have a good reason to please.

In this case it doesn't matter, both are equally (heh) good, but some linters will whine at you if you don't use triple =.

lib/bridge.js
```diff
@@ -237,11 +240,109 @@ Bridge.prototype.run = function(port, config, appServiceInstance) {
...
+                res.status(404).send("Cannot GET " + req.path + "\r\n");
+                return;
+            }
+
+            if (!lookupController.getProtocol) {
```
Kegsay: Combine the two if-statements :)

```javascript
if (req.params.ver != "unstable" || !lookupController.getProtocol) {
    res.status(404).send("Cannot GET " + req.path + "\r\n");
    return;
}
```

lib/bridge.js
```diff
@@ -237,11 +240,109 @@ Bridge.prototype.run = function(port, config, appServiceInstance) {
...
+            var protocol = req.params.protocol;
+
+            if (protocolFilter && !(protocol in protocolFilter)) {
+                res.status(404).json({err: "Unknown 3PN protocol " + protocol});
+                return;
+            }
+
+            lookupController.getProtocol(protocol).then(
+                (result) => { res.status(200).json(result) },
+                (e)      => { res.status(e.code).json({error: e.err}) }
```
Kegsay: No ES6 fat arrow functions please - ES5 only! Also, Please stick to existing whitespace styles: so no alignment on functions like this please (I admit it looks nice though :)).

lib/bridge.js
```diff
@@ -237,11 +240,109 @@ Bridge.prototype.run = function(port, config, appServiceInstance) {
...
+            var protocol = req.params.protocol;
+
+            if (protocolFilter && !(protocol in protocolFilter)) {
+                res.status(404).json({err: "Unknown 3PN protocol " + protocol});
+                return;
+            }
+
+            lookupController.getLocation(protocol, req.query).then(
+                (result) => { res.status(200).json([result]) },
+                (e)      => { res.status(e.code).json({error: e.err}) }
```
Kegsay: We need to be more paranoid. What if the error `e` does not contain a `code`? We might fall over in spectacular fashion. We should do some validation on the return result before blindly passing it through.

Kegsay:
Tests are failing due to use of ES6 arrow functions, which I've already commented on. Broadly speaking though, this is a good PR, thank you. :) Ping me when the requisite changes are done.

lib/bridge.js
leonerd: The fields are defined by whatever makes sense to each protocol individually. The choice of `protocol` effectively decides what fields are valid - that's the metadata returned by `getProtocol`.

lib/bridge.js
leonerd: They do yes - seems I forgot to document those. I'll add something.

lib/bridge.js
leonerd: I primarily split the code out to a new method just to keep it neater contained, away from the man `run` method. There's no pressing concern to document it and make it visible to subclasses. If that makes it awkward, maybe best just to not bother and keep it for internal purposes.

lib/bridge.js
leonerd: Fixed e68e083

lib/bridge.js
leonerd: Fixed e68e083

lib/bridge.js
leonerd: Done 099aa94

lib/bridge.js
leonerd: Done 94fa2cc

lib/bridge.js
leonerd: Done ea9b88e

lib/bridge.js
leonerd: Expanded docs ea7644c

lib/bridge.js
leonerd: Have added words to that effect in ea7644c

lib/bridge.js
Kegsay: The IRC bridge currently does something similar to this in order to do linking/unlinking of rooms for plumbing. If this were part of the public API, we could do it properly without having to gut wrench like we are currently doing (and will do in all other bridges like the Slack bridge).

Kegsay:
Two more things then I think we're good:
 - The return types nor the documentation mention that we expect `getProtocol`, `getLocation` and `getUser` to return **Promises** which resolve to the objects you've documented. We don't seem to support them just returning the object synchronously either. One or two extra words to say something to the effect of "return a promise that resolves to $OBJECT" would be great please. These docs are important because they will be converted to HTML and put up on gh-pages, which is why I'm fussing :)
 - https://github.com/matrix-org/matrix-appservice-bridge/pull/15#discussion_r77174536 isn't addressed yet.

lib/bridge.js
leonerd: OK; I've written an attempt at this in cc6a0a2.

I wonder if the function ought to handle the `/_matrix/app/:version` parts of the path itself though, to relieve call the callers from always having to remember to do it. Synapse has a similar structure, in which you even specify which API version(s) you support at the time you install the handler, so such version checking is done for you. This might look like (to pick some random example):

```
this.addAppServicePath({
    method: "GET",
    versions: ["r0", "r0.0.1"],
    path: "/thirdparty/throwAParty/:protocol",
    handler: function(req, res) { ... },
});
```

Generally, named arguments are nice, right?

lib/bridge.js
leonerd: Done df3c8c9

leonerd:
Oops, I think comment 15 got lost in the refactoring. Have done that now.
Have documented Promisey return values - 682182d

Kegsay:
LGTM

PR #17, opened 2016-09-05
Provide an Intent method for getProfileInfo()

leonerd:
This would be useful in the slack bridge:
  https://github.com/matrix-org/matrix-appservice-slack/blob/master/lib/BridgedRoom.js#L38


Kegsay:
ok to test

Kegsay:
test

Kegsay:
LGTM

PR #20, opened 2016-09-07
Make sure to .bind(this) to the thirdPartyLookup controller functions

leonerd:
This change is quite small but re-indents the function, so you might want to review it with `?s=0`


leonerd:
Actually thinking about it, this isn't helpful. That's binding the wrong `this`; there's a different one I want.

PR #22, opened 2016-09-12
ConfigValidator throws validate error with validation errors

lukebarnard1:
Validation errors thrown by ConfigValidator have a ```_validationErrors``` property which is set to the internal errors that occurred during validation, giving a more detailed report as to what failed to validate. This is useful for showing errors to users inputting data.


leonerd:
LGTM

PR #24, opened 2016-09-15
Implement reverse 3PL and 3PU lookups

leonerd:


lib/bridge.js
```diff
@@ -312,6 +312,28 @@ Bridge.prototype._customiseAppservice = function() {
 
         this.addAppServicePath({
             method: "GET",
+            path: "/_matrix/app/:ver/thirdparty/location",
+            handler: function(req, res) {
+                if (req.params.ver !== "unstable" || !lookupController.parseLocation) {
```
lukebarnard1: This needs a comment I think, not obvious that ```parseLocation``` is a function.

lib/bridge.js
```diff
@@ -312,6 +312,28 @@ Bridge.prototype._customiseAppservice = function() {
...
+                    res.status(404).send("Cannot GET " + req.path + "\r\n");
```
lukebarnard1: This error message isn't very useful

lib/bridge.js
```diff
@@ -332,6 +354,28 @@ Bridge.prototype._customiseAppservice = function() {
                 );
             }
         });
+
+        this.addAppServicePath({
+            method: "GET",
+            path: "/_matrix/app/:ver/thirdparty/user",
+            handler: function(req, res) {
+                if (req.params.ver !== "unstable" || !lookupController.parseUser) {
+                    res.status(404).send("Cannot GET " + req.path + "\r\n");
```
lukebarnard1: As above

lib/bridge.js
leonerd: Ah this one. @Kegsay commented on that too in the original functions.
That is attempting to preserve the same error message that `express` generates if it receives a request to a path it has no routes for.

lukebarnard1: But it's still not useful, someone might misinterpret that as "GET" is not valid or any part of the path  is invalid. Why not, "version not supported" or "function not supported" (for the ```!lookupController.parseLocation``` bit.

leonerd: Hmyes; perhaps the version check should be separated out into a different return value. I'll pop that (into all five)

leonerd:
PTAL - made something of a reworking of it. Each `app.get()` route is now only installed if there's a controller function for it anyway, so we don't have to test and return the 404 any more.
Additionally, I moved all the 3PE lookup logic into its own function so as to reduce the ever-growing indentation level here.

lib/bridge.js
leonerd: Now done

lib/bridge.js
leonerd: Hopefully now addressed by indirecting it via a variable called `parseLocationFunc`

PR #28, opened 2016-10-03
Add intent.sendReadReceipt()

leonerd:
I'd like bridges to send read receipts for the times when we get that information from the third-party network (gitter and slack for example).


Kegsay:
LGTM

Issue #29, opened 2016-10-04
{Slack,Gitter} bridge refactoring hitlist


leonerd:
This is less of a bug and more of a list of things (for my own sanity if nothing else) to extract from the current near copy-pasta mess of `matrix-appservice-{slack,gitter}` and put in the bridge library.

Entire files that can be copied (near-)verbatim
- [ ] `lib/AdminCommand.js`
- [ ] `lib/MatrixIdTemplate.js` from `matrix-appservice-gitter`
- [ ] `lib/MatrixUser.js`
- [x] `lib/Metrics.js`
- [ ] `lib/Provisioning.js`

Files that vary in implementation between Slack and Gitter but could share a common base class:
- [ ] `lib/BridgedRoom.js`
- [ ] `lib/SlackGhost.js` and `lib/GitterUser.js`

Then finally we come to the big one; all of the various parts of `lib/MatrixSlackBridge.js` and `lib/MatrixGitterBridge.js` that can be pulled out:
- [ ] automatic setup and use of `StateLookup`
- [ ] setup and delegation methods for metrics counters
- [x] automatic counting of bridge "intent" objects
- [ ] storage/metric counting/lazy construction of MatrixUser and {remote}User objects
- [ ] storage/metric counting/lazy construction of BridgedRoom instances, keyed bidirectionally
- [x] counting of Matrix client API calls
- [ ] returning a full `http` URL on the homeserver for a given `mxc`
- [ ] utility methods like `listGhostUsers`, `drainAndLeaveMatrixRoom`, `listRoomsFor`
- [ ] Matrix message routing and handling for admin room
- [ ] Checking user powerlevels for permission to execute a provisioning request

Half-Shot:
I think we're not hitting much of this list :)

PR #30, opened 2016-10-04
Pull `AdminCommand.js` from {Gitter,Slack,Telegram} bridges

leonerd:
The very initial start of #29.

An "AdminCommand" helps implement the "admin console" in the gitter and slack bridges - a special matrix room where the bridge owner can control the running of the bridge. (Not to be confused with the per-user admin rooms that the IRC bridge creates).


lib/components/admin-command.js
```diff
@@ -0,0 +1,157 @@
+"use strict";
+
+var minimist = require("minimist");
+
+/**
+ * Construct a new AdminCommand helper instance.
```
Kegsay: More detail please. Explain what this component does.

lib/components/admin-command.js
```diff
@@ -0,0 +1,157 @@
...
+ * @constructor
+ * @param {string} opts.desc A short description to display in the 'help'
```
Kegsay: I *think* you also need:
```js
/**
 * @param {Object} opts Options object
 */
```

for this to be valid JSDoc3.

lib/components/admin-command.js
```diff
@@ -0,0 +1,157 @@
...
+ * of the option, and the value is a string giving some descriptive text for
+ * the 'help' command.
+ * Each option name may be suffixed by aliases (for example, single letter
+ * short forms) separated by pipe symbols ('|'). Users may enter options in
+ * any of these shortcuts; the value presented to the command handler function
+ * will come in a key named after the first name component.
+ * Each option may be declared as mandatory, by prefixing the name string with
+ * an exclamation point ('!'). It then becomes an error to try to invoke the
+ * command without providing a value for that option; effectively turning it
+ * into a named argument instead.
```
Kegsay: I don't like this because it is overly complicated for no real reason. It sounds like what shape this should really be is:

```js
{
  $CMD_NAME: {
    required: true,
    description: "Some description",
    aliases: [$ALIAS, $ALIAS2]
  }
}
```

But instead we have to parse out what the caller gives us to form the equivalent object. Why aren't we just making this the API? It's clearer and involves writing less code.

lib/components/admin-command.js
```diff
@@ -0,0 +1,157 @@
...
+                    respond("  --" + n + ": " + optspec[n].desc);
+                });
+            }
+        },
+    });
+};
+
+/**
+ * A specification of command-line options recognised by an AdminCommand.
+ * @typedef {Object} AdminCommand~opts
```
Kegsay: Naming: I would prefer a less generic name here than `opts` (which is made worse by the fact that this class accepts `opts` in its constructor!). You say this is a "specification", so why not `spec`, `schema`, or better yet, identify what it is specifying, in this case commands e.g. `commandSchema` or `cmdSpec`.

lib/components/admin-command.js
```diff
@@ -0,0 +1,157 @@
...
+        // Canonicalise aliases
+        Object.keys(this.optaliases).forEach((a) => {
+            if (a in opts) {
+                opts[this.optaliases[a]] = opts[a];
+                delete opts[a];
+            }
+        });
+
+        Object.keys(opts).forEach((n) => {
+            if (n === "_") return;
```
Kegsay: What does `_` mean? Why are we ignoring it?

lib/components/admin-command.js
```diff
@@ -0,0 +1,157 @@
...
+
+            if (!(n in this.optspec)) {
```
Kegsay: `if (!this.optspec[n])` is more typical, but they are equiv.

lib/components/admin-command.js
```diff
@@ -0,0 +1,157 @@
...
+ * options recognised by the command.
+ * @param {AdminCommand~func} opts.func The handler function that implements
+ * the command.
+ */
+function AdminCommand(opts) {
+    this.desc = opts.desc;
+    this._func = opts.func;
+
+    if (opts.opts) {
+        this.optspec = {};
```
Kegsay: Your intention here is to make `optspec` and `optaliases` be maps. Sadly, this is Javascript and `Objects` are poor map replacements because of this:
```js
> s = {}
{}
> s["hasOwnProperty"]
[Function: hasOwnProperty]
> "hasOwnProperty" in s
true
```

The short-hand ` = {}` will use the default `Object.prototype` which has a bunch of keys already inserted into your Map! You can get around this by *explicitly* creating an object without a prototype:

```js
> s = Object.create(null);
{}
> s["hasOwnProperty"]
undefined
> "hasOwnProperty" in s
false
```

This is particularly important when you don't control what the key names are called (as in this case).

lib/components/admin-command.js
```diff
@@ -0,0 +1,157 @@
...
+
+/**
+ * The handler function for an AdminCommand that contains its actual
+ * implementation code.
+ * @typedef {function} AdminCommand~func
+ * @param {Object} bridge The top-level bridge instance.
+ * @param {Object} opts Values of any commandline options the user supplied
+ * @param {string[]} args Remaining positional commandline arguments after the
+ * options have been parsed out.
+ * @param {function} respond The responder function that the command handler
```
Kegsay: In no place do we specify what this looks like - how is it invoked? Does it have any return args?

lib/components/admin-command.js
```diff
@@ -0,0 +1,157 @@
...
+            if (n === "_") return;
+
+            if (!(n in this.optspec)) {
+                throw Error("Unrecognised option: " + n);
+            }
+        });
+
+        var missing = [];
+        Object.keys(this.optspec).sort().forEach((n) => {
+            if (this.optspec[n].required && !(n in opts)) missing.push(n);
```
Kegsay: Linting may whine at you for not expanding this out with `{ }`.

lib/components/admin-command.js
```diff
@@ -0,0 +1,157 @@
...
+        });
+
+        if (missing.length) {
+            throw Error("Missing required options: " + missing.join(", "));
+        }
+
+        args = opts._;
+        delete opts["_"];
```
Kegsay: This looks like fun, what does it do?

lib/components/admin-command.js
```diff
@@ -0,0 +1,157 @@
...
+ * Helper method to construct a 'help' AdminCommand instance
+ * @param {Object} commands The object containing the application's
+ * AdminCommand instances.
+ * @return {AdminCommand} A new AdminCommand.
+ *
+ * Typically this method would be invoked just after the admin command object
+ * is declared, to provide the automatic online help command.
+ *
+ *   var commands = {};
+ *   commands.help = AdminCommand.makeHelpCommand(commands);
```
Kegsay: \o/ for example usage! See [the JS SDK docs](http://matrix-org.github.io/matrix-js-sdk/0.5.5/client.js.html#line2890) for the JSDoc syntax so it gets printed prettily as HTML.

lib/components/admin-command.js
```diff
@@ -0,0 +1,157 @@
...
+                var name = args.shift();
+                var cmd = commands[name];
+                if (!cmd) {
+                    throw Error("No such command '" + name + "'");
+                }
+
+                respond(name + " - " + cmd.desc);
+                var optspec = cmd.optspec || {};
+                Object.keys(optspec).sort().forEach((n) => {
+                    respond("  --" + n + ": " + optspec[n].desc);
```
Kegsay: I'm not sure how I feel about this `respond` stuff. I assume this is a line-by-line thing, but I dislike it because the naive form is error prone:

```js
function respond(someString) {
  doHttpHitWithMessage(someString).then(function(res) {
    // yay it worked
  }, function(err) {
    // sadsies it no worky.
  });
}
```

Because `doHttpHitWithMessage` is async, this will race and you might get out-of-order messages. If we expect the `respond` implementation to be sending stuff to Matrix, this also will be grossly counting against our rate limiting count, whereas a `\n` message wouldn't.

lib/components/admin-command.js
```diff
@@ -0,0 +1,157 @@
...
+
+        if (missing.length) {
+            throw Error("Missing required options: " + missing.join(", "));
+        }
+
+        args = opts._;
+        delete opts["_"];
+    }
+
+    return this._func(bridge, opts, args, respond);
```
Kegsay: Right, now I realise what felt wrong about the API shape. We're passing functions around, which in and of itself is fine, but we **never do any async operations**. Why is `respond` A Thing? At a high level, I can see the API shape being:

```js
constructor(opts) {
  // determine your valid flags and descriptions etc
}

process(rawString) {
 // parse string into arguments
 // match them against the command schema in the constructor
 // return them to the caller
}
```

This would make the usage pretty clear too:

```js
var ac = new AdminCommand({
  "verbose": {
    required: false,
    description: "Enable verbose mode",
    aliases: ["v"]
  },
  "url": {
    required: true,
    description: "HTTP URL to hit",
    aliases: ["u"]
  }
});

var cmd = ac.process("--url https://matrix.org -v");
if (cmd.verbose) {
  // turn on verbose mode
}
console.log("URL is %s", cmd.url);
```

This also feels like it should've been done before though..

Kegsay:
I'm not convinced that this component is required. It feels like a combination of `minimist` to parse a raw string into `argv` and [Commander](https://www.npmjs.com/package/commander) would be performing much the same operations as this component. Am I missing something subtle here?

leonerd:
I'll reply first to the main comment, before tackling some of the individual ones.

I hadn't previously been aware of Commander, but looking at it it appears to be a way to build an entire self-contained program that runs, parses the actual program's argv, does one thing, then exits. That's not what this is for.

This class is for running commands on the "admin console", the special Matrix room that the bridge maintainer(s) join to control the regular running of the bridge. I did try several commandline wrapping modules with that, but most of them suffer too much "leakage", like printing to the real stderr on errors, or causing the entire process to exit.

lib/components/admin-command.js
leonerd: That's a minimist thing - `_` is the special key that contains a list of the remaining positional parameters it didn't think matched any named options.

lib/components/admin-command.js
leonerd: Well, only if there's never any present-but-false values, though in my case there aren't.

lib/components/admin-command.js
leonerd: (as from above): Extracts the remaining positional parameters that minimist has put into `_`.

lib/components/admin-command.js
leonerd: Ah; for this it might be best to see further uses of it, from the slack/gitter bridges. In summary, the `respond` function simply appends to a string buffer, to be emitted into matrix in a single message once the entire command has finished. It just found that neater to write code for than having the commands all return a single response string at the end

leonerd:
For example: these are creations of it:
https://github.com/matrix-org/matrix-appservice-slack/blob/master/lib/AdminCommands.js

which are then invoked by
https://github.com/matrix-org/matrix-appservice-slack/blob/master/lib/MatrixSlackBridge.js#L457-L504

lib/components/admin-command.js
leonerd: Done 8ad82d0

lib/components/admin-command.js
```diff
@@ -0,0 +1,157 @@
...
+ * @param {string} opts.desc A short description to display in the 'help'
+ * command.
+ * @param {AdminCommand~opts=} opts.opts An optional object giving commandline
+ * options recognised by the command.
+ * @param {AdminCommand~func} opts.func The handler function that implements
+ * the command.
+ */
+function AdminCommand(opts) {
+    this.desc = opts.desc;
+    this._func = opts.func;
```
leonerd: Done 1a4408e

lib/components/admin-command.js
leonerd: Done 2fe3412

lib/components/admin-command.js
leonerd: Done f437c6a

lib/components/admin-command.js
leonerd: I had a good long stare around those docs looking for how to make example code and I gave up trying to find it. Thanks.

Done 1c01451

Kegsay:
> most of them suffer too much "leakage", like printing to the real stderr on errors, or causing the entire process to exit.

100% agreed that most of them do indeed do stuff like this. I *think* Commander doesn't leak like this though (but I haven't used it). It just seems rather feature creepy of us to bundle this in, which we then have to maintain going forwards.

lib/components/admin-command.js
leonerd: Refactored by 31cc2a8 to extract earlier + commented.

lib/components/admin-command.js
Kegsay: Yes, it is longer. It is also **infinitely clearer** what the arguments are. Clarity > All.

kfatehi:
Is this for a combination "protocol room" and "command processor" ? 

E.g. in matrix-puppet-bridge we create a protocol status room (e.g. facebook status room) where we send messages like errors, connection info, etc, and join ghost users into. and there is also a command processor capability described here https://github.com/matrix-hacks/matrix-puppet-bridge/blob/master/README.md#q-how-can-i-add-bang--commands-to-a-room-such-as-echo although it hasn't really been used.

I'm trying to understand if this component is in essence going to allow me to replace those two things with the above? i.e. it is a way to have it automatically create/manage a protocol status room in addition to providing a way to define a set of commands that the user can invoke by typing into that room. is that correct?

leonerd:
PTAL.

Newly rewritten to a more expressive input form of objects with named keys, rather than parsing multiple things out of the string forms earlier. This is now common across the Gitter, Slack and Telegram ASes.

Half-Shot:
I don't think this is likely to be merged at this point, as a lot of stuff has changed (for instance, the Slack Bridge is moving onto Typescript and uses a third party lib to parse commands)

PR #32, opened 2016-10-18
Upport `Metrics.js` out of matrix-appservice-{slack,gitter}

leonerd:
Another step along #29.

I'm hoping after this is in, to pull some more of the integration code out of the specific bridges and into the toplevel bridge object too, so it can export some gauges automatically and also apply hit counters to bits of internal API I can't see easily from the outside.


leonerd:
I haven't added any `@example`s yet because I'm not sure where to put them. I want to show how to use `addGauge` with and without a refresh function, and how to use the counters.

lib/components/agecounters.js
```diff
@@ -0,0 +1,57 @@
...
+ * hourly/daily/weekly active instances of various types of object within the
+ * bridge.
+ *
+ * @constructor
+ */
+function AgeCounters() {
+    this["1h"] = 0;
+    this["1d"] = 0;
+    this["7d"] = 0;
+    this["all"] = 0;
```
Kegsay: What is the intention behind assigning counters like this as opposed to:
```js
function AgeCounters() {
  this.counters = {
    "1h": 0,
    "1d": 0,
    "7d": 0,
    all: 0,
  }
}
```
I care because in its current form, `:48` is rather error prone if we ever add **any** new property to `AgeCounters`. If we namespaced it all under `counters`, this would make `Object.keys(this)` much less error-prone because it would be replaced with `Object.keys(this.counters)`, meaning new properties wouldn't clash.

lib/components/agecounters.js
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Fetch the counts in the age buckets and set them as labeled observations in
+ * the given gauge metric instance.
+ *
+ * @param {Gauge} gauge The gauge metric instance.
+ * @param {Object} morelabels An object containing more labels to add to the
+ * gauge when setting values.
+ */
+AgeCounters.prototype.setGauge = function(gauge, morelabels) {
+    Object.keys(this).forEach((age) => {
```
Kegsay: Please do not use fat arrows. This library needs to work on ES5.

lib/components/agecounters.js
```diff
@@ -0,0 +1,57 @@
...
+        // I wish I could use spread expressions
+        var labels = {age: age};
+        Object.keys(morelabels).forEach((k) => labels[k] = morelabels[k]);
```
Kegsay: Please do not use fat arrows. This library needs to work on ES5.

lib/components/metrics.js
```diff
@@ -0,0 +1,293 @@
...
+ *   function makeThing() {
+ *       metrics.incCounter("things_made");
+ *       return new Thing();
+ *   }
+ *
+ * @constructor
+ */
+function Metrics() {
+    // Only attempt to load these dependencies if metrics are enabled
+    var Prometheus = require("prometheus-client");
```
Kegsay: This should be mentioned in the class docs above.

lib/components/metrics.js
```diff
@@ -0,0 +1,293 @@
...
+
+    var client = this._client = new Prometheus();
+
+    this._collectors = []; // executed in order
+    this._counters = {};
```
Kegsay: What are the keys? Counters? Values? Counters? Inline comment the schema here please.

lib/components/metrics.js
```diff
@@ -0,0 +1,293 @@
...
+ *   });
+ *
+ *   function makeThing() {
+ *       metrics.incCounter("things_made");
+ *       return new Thing();
+ *   }
+ *
+ * @constructor
+ */
+function Metrics() {
```
Kegsay: Given the strong dependency on Prometheus, I would prefer if this was called `PrometheusMetrics`. The de facto metrics tool in Node is Statsd, so people may naively think that this is actually a statsd metrics class.

lib/components/metrics.js
```diff
@@ -0,0 +1,293 @@
...
+
+    var heap_size_gauge = this.addGauge({
+        namespace: "process",
+        name: "heap_bytes",
+        help: "Total size of Node.js heap in bytes",
+    });
+    var heap_used_gauge = this.addGauge({
+        namespace: "nodejs",
+        name: "heap_used_bytes",
+        help: "Used size of Node.js heap in bytes",
```
Kegsay: This is all wonderful :heart: 

Why don't we [do something similar](https://github.com/matrix-org/matrix-appservice-bridge/pull/30#discussion_r83233542) with `AdminCommands`? :angel:

lib/components/metrics.js
```diff
@@ -0,0 +1,293 @@
...
+        // Virtual memory size
+        vsz_gauge.set({}, stats[20]);
+    });
+
+    this.addGauge({
+        namespace: "process",
+        name: "open_fds",
+        help: "Number of open file descriptors",
+        refresh: function(gauge) {
+            var fds = fs.readdirSync("/proc/self/fd");
```
Kegsay: Why are we actively blocking the event loop whilst reading this? Why aren't we using the async form?

lib/components/metrics.js
```diff
@@ -0,0 +1,293 @@
...
+        namespace: "process",
+        name: "start_time_seconds",
+        help: "Start time of the process since unix epoch in seconds",
+    }).set({}, _calculate_process_start_time());
+
+    this.refresh();
+};
+
+function _read_proc_self_stat() {
+    var stat_line = fs.readFileSync("/proc/self/stat")
```
Kegsay: Why are we actively blocking the event loop whilst reading this? Why aren't we using the async form?

lib/components/metrics.js
```diff
@@ -0,0 +1,293 @@
...
+            gauge.set(null, fds.length - 1);
+        }
+    });
+
+    this.addGauge({
+        namespace: "process",
+        name: "max_fds",
+        help: "Maximum number of open file descriptors allowed",
+        refresh: function(gauge) {
+            var limits = fs.readFileSync("/proc/self/limits");
```
Kegsay: Why are we actively blocking the event loop whilst reading this? Why aren't we using the async form?

lib/components/metrics.js
```diff
@@ -0,0 +1,293 @@
...
+
+    var btime_line = fs.readFileSync("/proc/stat")
+        .toString().split(/\n/).filter((l) => l.match(/^btime /))[0];
+    var btime = Number(btime_line.split(" ")[1]);
+
+    return btime + starttime_sec;
+}
+
+Metrics.prototype.refresh = function() {
+    this._collectors.forEach((f) => f());
```
Kegsay: The function being called here does a lot of UNIX specific stuff which will just die outright on Windows. We should be `try {} catch {}`ing on this function call.

Also, please do not use fat arrows, this needs to work on ES5.

lib/components/metrics.js
```diff
@@ -0,0 +1,293 @@
...
+    var refresh = opts.refresh;
+    var gauge = this._client.newGauge({
+        namespace: opts.namespace || "bridge",
+        name: opts.name,
+        help: opts.help,
+    });
+
+    if (opts.refresh) {
+        var refresh = opts.refresh;
+        this._collectors.push(() => refresh(gauge));
```
Kegsay: Try/catch and fat arrows. Same as https://github.com/matrix-org/matrix-appservice-bridge/pull/32/files#r83883638

lib/components/metrics.js
```diff
@@ -0,0 +1,293 @@
...
+Metrics.prototype.addGauge = function(opts) {
+    var refresh = opts.refresh;
+    var gauge = this._client.newGauge({
+        namespace: opts.namespace || "bridge",
+        name: opts.name,
+        help: opts.help,
+    });
+
+    if (opts.refresh) {
+        var refresh = opts.refresh;
```
Kegsay: Redeclaration from `:228`.

lib/components/metrics.js
```diff
@@ -0,0 +1,293 @@
...
+};
+
+/**
+ * Increments the value of a counter metric
+ * @param{string} name The name the metric was previously registered as.
+ * @param{Object} labels Optional object containing additional label values.
+ */
+Metrics.prototype.incCounter = function(name, labels) {
+    if (!this._counters[name]) {
+        console.log("TODO: missing metric " + name);
```
Kegsay: Feels like this should throw?

lib/components/metrics.js
```diff
@@ -0,0 +1,293 @@
...
+ * containing Express app.
+ * @param {Bridge} bridge The containing Bridge instance.
+ */
+Metrics.prototype.addAppServicePath = function(bridge) {
+    var metricsFunc = this._client.metricsFunc();
+
+    bridge.addAppServicePath({
+        method: "GET",
+        path: "/metrics",
+        handler: (req, res) => {
```
Kegsay: No fat arrows please.

lib/components/metrics.js
```diff
@@ -0,0 +1,293 @@
...
+ * Adds a new counter metric
+ * @param {Object} opts Options
+ * @param {string} opts.namespace An optional toplevel namespace name for the
+ * new metric. Default: <code>"bridge"</code>.
+ * @param {string} opts.name The variable name for the new metric.
+ * @param {string} opts.help Descriptive help text for the new metric.
+ * Once created, the value of this metric can be incremented with the
+ * <code>incCounter</code> method.
+ */
+Metrics.prototype.addCounter = function(opts) {
```
Kegsay: `addCounter` feels like this is adding to a list, but in reality it will clobber based on the `opts.name` provided. This isn't made clear in the JSDoc.

Kegsay:
Main complaints are:
 - Not considering ES5 support (the rest of this library is ES5-compliant)
 - Not considering Windows support (no suitable try/catch on accessing UNIX specific stuff, also not enough paranoia on the values contained inside said files)
 - Accessing files synchronously (this is a dealbreaker for me, I cannot afford to stop the world in the IRC bridge)

Other than that, this looks very useful as a standard metrics package for the bridges :D

lib/components/agecounters.js
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Increment the values of the internal counters depending on the given age,
+ * in seconds.
+ *
+ * @param {Number} age The age in seconds.
+ */
+AgeCounters.prototype.bump = function(age) {
+    if (age < HOUR   ) this["1h"]++;
+    if (age < DAY    ) this["1d"]++;
+    if (age < DAY * 7) this["7d"]++;
```
leonerd: Huh. notsure. Done 89aaf09

lib/components/agecounters.js
leonerd: Done 9fbdaaf

lib/components/agecounters.js
leonerd: Done 9fbdaaf

lib/components/metrics.js
leonerd: Done ef77732 (and elsewhere)

lib/components/metrics.js
leonerd: Done 5fff9b2

lib/components/metrics.js
leonerd: That sounds good. I'll rename it last though to avoid upsetting github too much.

lib/components/metrics.js
leonerd: (see answer below)

lib/components/metrics.js
```diff
@@ -0,0 +1,293 @@
...
+    return stat_line.match(/\) +(.*)$/)[1].split(" ");
+}
+
+function _calculate_process_start_time() {
+    // The 'starttime' field in /proc/self/stat gives the number of CPU ticks
+    //   since machine boot time that this process started.
+    var stats = _read_proc_self_stat();
+    var starttime_sec = stats[19] / TICKS_PER_SEC;
+
+    var btime_line = fs.readFileSync("/proc/stat")
```
leonerd: (see answer above)

lib/components/metrics.js
leonerd: Removed 5793d57

lib/components/metrics.js
```diff
@@ -0,0 +1,293 @@
...
+ * It runs purely to have a side-effect on previously registered gauges.
+ */
+Metrics.prototype.addCollector = function(func) {
+    this._collectors.push(func);
+};
+
+/**
+ * Adds a new gauge metric.
+ * @param {Object} opts Options
+ * @param {string} opts.namespace An optional toplevel namespace name for the
```
leonerd: Done 9f8afb8

lib/components/metrics.js
leonerd: Ohyes, that was left over from gradually introducing the code into the gitter/slack bridges.

Now throws ab806bf

lib/components/metrics.js
leonerd: Done 543a998

lib/components/metrics.js
leonerd: The prometheus client itself already enforces uniqueness of names among the metrics, so in actual fact both this and `addGauge` will throw on attempts to add a second copy of the same name. Should I document that fact?

lib/components/metrics.js
leonerd: I think rather than `try {} catch {}` wrapping the collector functions, better would be to only even enable those metrics if `/proc/self/stat` is found to be working at construction time.

Done in 84b5cb7

lib/components/metrics.js
leonerd: Renamed 9ae9dcd

leonerd:
> Not considering ES5 support (the rest of this library is ES5-compliant)
I believe this is now solved by not using `=>` syntax.

> Not considering Windows support (no suitable try/catch on accessing UNIX specific stuff, also not enough paranoia on the values contained inside said files)
Process-wide metrics are now disabled if `/proc/self/stat` isn't found to be working at construction time.

> Accessing files synchronously (this is a dealbreaker for me, I cannot afford to stop the world in the IRC bridge)
See comments above about `/proc` filesystem being effectively synchronous anyway.

PTAL @Kegsay

lib/components/prometheusmetrics.js
```diff
@@ -0,0 +1,315 @@
...
+    var Prometheus = require("prometheus-client");
+
+    var client = this._client = new Prometheus();
+
+    this._collectors = []; // executed in order
+    this._counters = {}; // counter metrics keyed by name
+
+    var have_proc_stat = false;
+    try {
+        fs.readFileSync("/proc/self/stat");
```
Kegsay: > Whlie this is a Sync form of call, using the async form won't actually make any difference. The /proc/self/ introspection API in the kernel is implemented in an interesting way that doesn't play nicely with aio, so libuv's async versions of file reading won't be any different on those.
In any case, the /proc entries won't block on reading; they're designed to be fast and lightweight to respond

Please can you put some version of said comment here to make it clear for any future readers.

Kegsay:
LGTM - just needs a comment about accessing files syncly.

lib/components/prometheusmetrics.js
leonerd: Done 086553a

leonerd:
commented.

Kegsay:
LGTM! :star:

PR #36, opened 2016-11-04
Begin hooking up PrometheusMetrics to core Bridge object

leonerd:
Part of #29 

 * Counting of MatrixClient API calls
 * Counting of intent users
 * Neater wrapping of other gauge counts to simplify bridge instance code

### DIFF
--- a/lib/bridge.js
+++ b/lib/bridge.js
@@ -683,6 +683,8 @@ Bridge.prototype._updateIntents = function(event) {
  * Returns a PrometheusMetrics instance stored on the bridge, creating it first
  * if required. The instance will be registered with the HTTP server so it can
  * serve the "/metrics" page in the usual way.
+ * The instance will automatically register the Matrix SDK metrics by calling
+ * {PrometheusMetrics~registerMatrixSdkMetrics}.
  */
 Bridge.prototype.getPrometheusMetrics = function() {
     if (this._metrics) {
@@ -710,6 +712,18 @@ Bridge.prototype.getPrometheusMetrics = function() {
  * convenience.
  * @param {PrometheusMetrics~BridgeGaugesCallback} counterFunc A function that
  * when invoked returns the current counts of various items in the bridge.
+ *
+ * @example
+ * bridge.registerBridgeGauges(() => {
+ *     return {
+ *         matrixRoomConfigs: Object.keys(this.matrixRooms).length,
+ *         remoteRoomConfigs: Object.keys(this.remoteRooms).length,
+ *
+ *         remoteGhosts: Object.keys(this.remoteGhosts).length,
+ *
+ *         ...
+ *     }
+ * })
  */
 Bridge.prototype.registerBridgeGauges = function(counterFunc) {
     var self = this;
--- a/lib/bridge.js
+++ b/lib/bridge.js
@@ -693,6 +693,27 @@ Bridge.prototype.getPrometheusMetrics = function() {
     return metrics;
 };
 
+/**
+ * A convenient shortcut to calling registerBridgeGauges() on the
+ * PrometheusMetrics instance directly. This version will supply the value of
+ * the matrixGhosts field if the counter function did not return it, for
+ * convenience.
+ * @param {PrometheusMetrics~BridgeGaugesCallback} counterFunc A function that
+ * when invoked returns the current counts of various items in the bridge.
+ */
+Bridge.prototype.registerBridgeGauges = function(counterFunc) {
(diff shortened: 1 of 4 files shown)


Kegsay:
ok to test

lib/bridge.js
```diff
@@ -672,6 +679,48 @@ Bridge.prototype._updateIntents = function(event) {
     });
 };
 
+/**
+ * Returns a PrometheusMetrics instance. The instance will be registered with
+ * the HTTP server so it can serve the "/metrics" page in the usual way.
+ */
+Bridge.prototype.getPrometheusMetrics = function() {
```
Kegsay: It's not clear that this will implicitly *create* a `PrometheusMetrics`. Please mention this in the docs.

Kegsay:
Linting errors need fixing.

lib/bridge.js
```diff
@@ -672,6 +679,48 @@ Bridge.prototype._updateIntents = function(event) {
...
+/**
+ * A convenient shortcut to calling registerBridgeGauges() on the
+ * PrometheusMetrics instance directly. This version will supply the value of
+ * the matrixGhosts field if the counter function did not return it, for
+ * convenience.
+ * @param {PrometheusMetrics~BridgeGaugesCallback} counterFunc A function that
+ * when invoked returns the current counts of various items in the bridge.
+ */
+Bridge.prototype.registerBridgeGauges = function(counterFunc) {
+    this.getPrometheusMetrics().registerBridgeGauges(() => {
```
Kegsay: I don't like us not returning a `Promise` to let the caller know when this gauge has been registered. Please return it. Also, ES5 only (no fat arrow functions!).

lib/components/prometheusmetrics.js
```diff
@@ -208,6 +208,122 @@ PrometheusMetrics.prototype._register_process_metrics = function() {
...
+     * of every time they're called. This seems to be neater than trying to
+     * intercept all HTTP requests and try to intuit what internal method was
+     * invoked based on the HTTP URL.
+     */
+
+    var matrixClientPrototype = require("matrix-js-sdk").MatrixClient.prototype;
+
+    var CLIENT_METHODS = [
+        "ban",
+        "createAlias",
```
Kegsay: Why are we doing this? I thought the plan was to hook into the underlying request function?

I don't like this in its current form because:
 - This list will get out of date with the JS SDK as new APIs are added or new convenience functions are added.
 - This list is a PITA to maintain.
 - This list does not cover the full scope of API calls that a bridge can do (e.g. it doesn't catch raw API calls the IRC bridge does to `/initialSync`).

I am not convinced we need this. Please convince me otherwise.

Kegsay:
This is getting sufficiently complex that I don't know how to use it. Can you document some examples somewhere please (e.g. the README)?

lib/bridge.js
leonerd: It's all instantaneous. It's all just in-memory state, there's no network IO involved at this time.
Fat arrow removed by 93698db

lib/bridge.js
leonerd: Doced b83ffd2

lib/components/prometheusmetrics.js
leonerd: That was my original plan, but then I hit a snag. The snag is in summary - REST.

The URLs visible at the http `request` function contain embedded in them lots of IDs - room IDs, user IDs, room state event types, etc... In order to unpick those and decode what the actual API endpoint is would require keeping a big list of all the various API endpoints with glob patterns or similar to unpick the IDs out of the URLs and then turn them into method names. E.g. dendron keeps such a list; this entire file is dedicated to such configuration:

https://github.com/matrix-org/dendron/blob/develop/src/github.com/matrix-org/dendron/proxy/paths.go

I decided that maintaining another list like that would be *even harder* than just keeping a list of SDK method names, because at least now I don't have to parse URLs or adapt to shape changes there, only names of methods within the MatrixClient SDK class itself.

lib/bridge.js
leonerd: Also e3f84aa

leonerd:
Where's a good place to add such an example? Inline in the main `bridge.js` docs?

Kegsay:
That or the README.

Kegsay:
Aside from docs, LGTM!

lib/components/prometheusmetrics.js
leonerd: Added 40eac45

leonerd:
Added example 169271c; PTAL

Kegsay:
LGTM

PR #37, opened 2016-11-05
Add remove by Id functions for room store.

Half-Shot:


lib/components/room-bridge-store.js
```diff
@@ -351,6 +351,20 @@ RoomBridgeStore.prototype.removeEntriesByRemoteRoomData = function(data) {
     return this.delete(data);
 };
 
+/**
+ * Remove entries with this remote room id.
+ * @param {Object} remoteId The remote id.
+ * @return {Promise}
+ * @example
+ *  new RemoteRoom("foobar");
```
lukebarnard1: Maybe one less space before 'new' (but +1 for `@example`, I didn't even know that was a thing)

lib/components/room-bridge-store.js
```diff
@@ -373,6 +387,21 @@ RoomBridgeStore.prototype.removeEntriesByMatrixRoomData = function(data) {
 };
 
 /**
+ * Remove entries with this matrix room id.
+ * @param {Object} matrixId The matrix id.
+ * @return {Promise}
+ * @example
+ *  new MatrixRoom("!foobar:matrix.org");
```
lukebarnard1: Same again, remove one space before `new`

PR #40, opened 2016-11-15
Add StateLookup.untrackRoom() method

### LINKED ISSUE #26: Please provide StateLookup.untrackRoom()
Currently I have some hideous monkeypatching:

https://github.com/matrix-org/matrix-appservice-gitter/commit/19aaa4cbfea35c445e37ebfa43f3126e0be8aa7f#diff-73f5f6969c2293ed1df0158153678a6aR23

leonerd:
Fixes #26 

See also
    https://github.com/matrix-org/matrix-appservice-slack/blob/master/lib/MatrixSlackBridge.js#L21
    https://github.com/matrix-org/matrix-appservice-gitter/blob/master/lib/Main.js#L25

### DIFF
--- a/lib/components/state-lookup.js
+++ b/lib/components/state-lookup.js
@@ -120,10 +120,10 @@ StateLookup.prototype.trackRoom = function(roomId) {
 /**
  * Stop tracking a given room.
  *
- * This will stop further tracking of state events in he given room and delete
+ * This will stop further tracking of state events in the given room and delete
  * existing stored state for it.
  *
- * @param{string} roomId The room ID to stop tracking.
+ * @param {string} roomId The room ID to stop tracking.
  */
 StateLookup.prototype.untrackRoom = function(roomId) {
     delete this._dict[roomId];
--- a/lib/components/state-lookup.js
+++ b/lib/components/state-lookup.js
@@ -99,7 +99,7 @@ StateLookup.prototype.trackRoom = function(roomId) {
                         r.events[ev.type][ev.state_key] = ev;
                     }
                 });
-                resolve();
+                resolve(r);
             }, function(err) {
                 if (err.httpStatus >= 400 && err.httpStatus < 600) { // 4xx, 5xx
                     reject(err); // don't have permission, don't retry.
@@ -148,7 +148,7 @@ StateLookup.prototype.onEvent = function(event) {
         // event and always use the pushed events we should remain in sync.
         // we'll add our own listener for the sync promise and then update this
         // value.
-        r.syncPromise.then(function() {
+        r.syncPromise.then(function(r) {
             if (!r.events[event.type]) {
                 r.events[event.type] = {};
             }
--- a/lib/components/state-lookup.js
+++ b/lib/components/state-lookup.js
@@ -149,6 +149,8 @@ StateLookup.prototype.onEvent = function(event) {
         // we'll add our own listener for the sync promise and then update this
         // value.
         r.syncPromise.then(function(r) {
+            // We get 'r' passed back in from the resolve() call so we don't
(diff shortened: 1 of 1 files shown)


lib/components/state-lookup.js
```diff
@@ -116,6 +118,18 @@ StateLookup.prototype.trackRoom = function(roomId) {
 };
 
 /**
+ * Stop tracking a given room.
+ *
+ * This will stop further tracking of state events in he given room and delete
```
Kegsay: `in he given `

lib/components/state-lookup.js
```diff
@@ -116,6 +118,18 @@ StateLookup.prototype.trackRoom = function(roomId) {
...
+ * existing stored state for it.
+ *
+ * @param{string} roomId The room ID to stop tracking.
```
Kegsay: Space after `@param` please.

lib/components/state-lookup.js
```diff
@@ -106,6 +106,8 @@ StateLookup.prototype.trackRoom = function(roomId) {
                 }
                 // wait a bit then try again
                 Promise.delay(3000).then(function() {
+                    if (!self._dict[roomId]) return;
```
Kegsay: ESLint may whine at you for this.

Kegsay:
Deleting stuff is Hard sadly.

This leaks in the following (common) scenario:
 - `trackRoom(roomId)` : Makes a `_dict` entry.
 - `onEvent` : Called before `syncPromise` is resolved so adds a resolved callback.
 - /sync succeeds, things resolve and and stuff happens.
 - `untrackRoom(roomId)` : Deletes `_dict` entry.
 - The deleted entry cannot be GCed because it is trapped in a closure from the resolved callback at `:149`.

This also leaks in the following rare scenario:
 - `trackRoom(roomId)` : Makes a `_dict` entry.
 - `onEvent` : Called before `syncPromise` is resolved so adds a resolved callback.
 - /sync does NOT succeed. Start waiting 3s.
 - `untrackRoom(roomId)` : Deletes `_dict` entry.
 - 3s timer expires. Never resolves/rejects the promise due to the null guard you added on `:109`.
 - The deleted entry cannot be GCed because it is trapped in a closure from the never-going-to-resolve callback at `:149`.

You'll need to re-architect this a bit if you actually want to delete state for realsies. Probably the simplest is to resolve *with the entry* since is doesn't touch anything outside (it doesn't *need* to wrap the entry in a closure) and then change it from being an anonymous function to being a member function to remove the closure. Alternatively, we shouldn't copy the reference to `r` and instead do what we do in `trackRoom` and constantly refer to (and null guard) `this._dict[roomId]`.

Kegsay:
Otherwise LGTM!

lib/components/state-lookup.js
leonerd: Fixed 6ec257e

lib/components/state-lookup.js
leonerd: Fixed 6ec257e

lib/components/state-lookup.js
leonerd: Fixed 6c32765

leonerd:
Yes - I generally prefer the pattern of not storing promises that implicitly capture an object inside that same object; passing it in as a resolve param would be nicer.

I believe that's now done in 100e17e - @Kegsay PTAL

Kegsay:
`ETOOMANYRs` - I think this works, but I'm not sure if you did that on purpose or not. It works because you're now variable shadowing in the resolved function, making it impossible for the resolve function to reference the closured `r`. I can only assume that V8 is smart enough to know that. A small comment to point this out would be good.

LGTM.

Issue #43, opened 2016-12-02
Switch prometheus code to `prom-client`


leonerd:
https://www.npmjs.com/package/prom-client

Seems to be

 * Better engineered
 * Already supports all the standard prometheus process-wide metrics, and a bunch of node-specific ones
 * Supports Histograms and Summaries

This latter feature is relevant to https://github.com/matrix-org/matrix-appservice-irc/issues/296#issuecomment-264514506

leonerd:
Started working on https://github.com/matrix-org/matrix-appservice-bridge/tree/paul/prom-client

leonerd:
Now running on `matrix-appservice-gitter`, seems to be fine.

I did have to upgrade to `node` version `v6.1.0` before the CPU usage metric started working. I then had to patch `prom-client` to report the user + system split metrics, rather than just the combined total.

See
https://github.com/matrix-org/prom-client/tree/paul/cpu-user-and-system

leonerd:
Done :)

PR #44, opened 2016-12-07
Switch from `prometheus-client` to `prom-client`

leonerd:
Main reason being that `prom-client` supports both Histograms and Summaries; old `prometheus-client` does neither.

Also, `prom-client` does its own implementation of the standard process-wide metrics, so we can delete *aaaall* the code relating to that here.

See also #43

### DIFF
--- a/lib/components/prometheusmetrics.js
+++ b/lib/components/prometheusmetrics.js
@@ -212,7 +212,7 @@ PrometheusMetrics.prototype.addCollector = function(func) {
  * new metric. Default: <code>"bridge"</code>.
  * @param {string} opts.name The variable name for the new metric.
  * @param {string} opts.help Descriptive help text for the new metric.
- * @param {string[]=} opts.labels An optional list of string label names
+ * @param {Array<string>=} opts.labels An optional list of string label names
  * @param {Function=} opts.refresh An optional function to invoke to generate a
  * new value for the gauge.
  * If a refresh function is provided, it is invoked with the gauge as its only
@@ -242,7 +242,7 @@ PrometheusMetrics.prototype.addGauge = function(opts) {
  * @param {string} opts.help Descriptive help text for the new metric.
  * Once created, the value of this metric can be incremented with the
  * <code>incCounter</code> method.
- * @param {string[]=} opts.labels An optional list of string label names
+ * @param {Array<string>=} opts.labels An optional list of string label names
  * @return {Counter} A counter metric.
  */
 PrometheusMetrics.prototype.addCounter = function(opts) {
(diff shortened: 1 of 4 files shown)


Kegsay:
ok to test

lib/components/prometheusmetrics.js
```diff
@@ -378,6 +212,7 @@ PrometheusMetrics.prototype.addCollector = function(func) {
  * new metric. Default: <code>"bridge"</code>.
  * @param {string} opts.name The variable name for the new metric.
  * @param {string} opts.help Descriptive help text for the new metric.
+ * @param {[string]=} opts.labels An optional list of string label names
```
Kegsay: Not valid JSDoc. `{string[]=}` please.

lib/components/prometheusmetrics.js
```diff
@@ -1,38 +1,11 @@
...
- *   process_virtual_memory_bytes
- *   process_heap_bytes
- *   process_cpu_seconds_total
- *   process_open_fds
- *   process_max_fds
- *   process_start_time_seconds
- *   nodejs_heap_used_bytes
- *
- * This class depends on having <code>prometheus-client</code> installed. It
+ * This class depends on having <code>prome-client</code> installed. It
```
Kegsay: `prome-client`

lib/components/prometheusmetrics.js
```diff
@@ -440,14 +309,26 @@ PrometheusMetrics.prototype.incCounter = function(name, labels) {
  * @param {Bridge} bridge The containing Bridge instance.
  */
 PrometheusMetrics.prototype.addAppServicePath = function(bridge) {
-    var metricsFunc = this._client.metricsFunc();
+    var register = this._client.register;
```
Kegsay: Surprised you don't need to `.bind(this)`.

Kegsay:
LGTM

lib/components/prometheusmetrics.js
```diff
@@ -431,7 +264,43 @@ PrometheusMetrics.prototype.incCounter = function(name, labels) {
...
+};
+
+/**
+ * Adds a new timer metric, represented by a prometheus Histogram.
+ * @param {Object} opts Options
+ * @param {string} opts.namespace An optional toplevel namespace name for the
+ * new metric. Default: <code>"bridge"</code>.
+ * @param {string} opts.name The variable name for the new metric.
+ * @param {string} opts.help Descriptive help text for the new metric.
+ * @param {[string]=} opts.labels An optional list of string label names
```
leonerd: fixed 2d43f21

lib/components/prometheusmetrics.js
leonerd: fixed 2d43f21

PR #49, opened 2017-02-02
Remove roomId param from _isRemoteUser

Half-Shot:
Not sure why this was a param, so I removed it.


Half-Shot:
Let's try that again with the right branch.

PR #51, opened 2017-03-29
Implement a 'dontJoin' intent option

leonerd:
If the bridge app itself is maintaining membership state then having the library set join events too might disrupt that. E.g. currently `matrix-appservice-tg` wants to keep a per-room displayname in membership events.


Kegsay:
LGTM

PR #54, opened 2017-05-05
Upgrade to Jasmine 2.5

Kegsay:
Required so we can mock out Dates correctly, which is needed for testing Intent sharing logic.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,9 +6,9 @@
   "scripts": {
     "gendoc": "jsdoc -r lib -R README.md -P package.json -d .jsdoc",
     "lint": "eslint lib/ spec/",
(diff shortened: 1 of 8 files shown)


lukebarnard1:
In theory this looks reasonable but the tests reveal some unhandled errors that _aren't_ causing them to fail.

Kegsay:
The unhandled errors are from turning BLUEBIRD_DEBUG=1 on which flags up any promises without rejection handlers. I'm not interested in making any functional changes to this already long PR. I suspect they are benign.

Kegsay:
They were benign, since they happen when:
```js
promise.finally(() => { ... })
```
is called, since there is no explicit error handling. As per [the docs](http://bluebirdjs.com/docs/api/suppressunhandledrejections.html) on suppressing unhandled rejection errors, I'm just attaching an empty `catch` block.

PR #55, opened 2017-05-05
Share Intent data structures created by Bridge

Kegsay:
This has 2 notable performance implications wrt the IRC bridge:
 - Prevents unbounded memory expansion as we now cull `Intent` objects.
 - Prevents CPU spins that happen when all `Intent` objects are looped over when updating `m.room.member` or `m.room.power_levels` events.

Requires #54 for Jasmine 2, so is based off that PR. With tests to assert that:
 - Culls happen.
 - Culling doesn't result in missing state when they get remade.

### DIFF
--- a/lib/bridge.js
+++ b/lib/bridge.js
@@ -18,6 +18,11 @@ var Promise = require("bluebird");
 var Datastore = require("nedb");
 var util = require("util");
 
+// The frequency at which we will check the list of accumulated Intent objects.
+var INTENT_CULL_CHECK_PERIOD_MS = 1000 * 60; // once per minute
+// How long a given Intent object can hang around unused for.
+var INTENT_CULL_EVICT_AFTER_MS = 1000 * 60 * 15; // 15 minutes
+
 /**
  * @constructor
  * @param {Object} opts Options to pass to the bridge
@@ -140,6 +145,9 @@ function Bridge(opts) {
         // user_id + request_id : Intent
     };
     this._intents["bot"] = null;
+    // _intent key => timestamp
+    this._intentLastAccessed = Object.create(null);
+    this._intentLastAccessedTimeout = null;
     this._membershipMap = {
         // room_id: { user_id: "join|invite|leave|ban|null" }   null=unknown
     };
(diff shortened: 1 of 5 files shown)


spec/integ/bridge.spec.js
```diff
@@ -500,6 +536,36 @@ describe("Bridge", function() {
             });
         });
 
+        it("should keep culled Intents up-to-date with incoming events", function(done) {
```
lukebarnard1: We'll have to go through this one together, I'm a bit confused as to why we shouldn't expect a join and what that has to do with keeping intents up-to-date

lib/bridge.js
```diff
@@ -18,6 +18,11 @@ var Promise = require("bluebird");
 var Datastore = require("nedb");
 var util = require("util");
 
+// The frequency at which we will check the list of accumulated Intent objects.
```
Kegsay: As per RL, people will understand either.

spec/integ/bridge.spec.js
Kegsay: Done. https://github.com/matrix-org/matrix-appservice-bridge/pull/55/commits/21a5f7478884e344631b8bb657da6a801401d3cd

lib/bridge.js
```diff
@@ -268,6 +289,29 @@ Bridge.prototype._customiseAppservice = function() {
     }
 };
 
+// Set a timer going which will periodically remove Intent objects to prevent
+// them from accumulating too much. Removal is based on access time (calls to
+// getIntent).
```
Kegsay: Done. https://github.com/matrix-org/matrix-appservice-bridge/pull/55/commits/21a5f7478884e344631b8bb657da6a801401d3cd

lib/bridge.js
```diff
@@ -140,6 +145,21 @@ function Bridge(opts) {
         // user_id + request_id : Intent
     };
     this._intents["bot"] = null;
+    // _intent key => timestamp
```
Kegsay: Done. https://github.com/matrix-org/matrix-appservice-bridge/pull/55/commits/21a5f7478884e344631b8bb657da6a801401d3cd

PR #58, opened 2017-05-18
Enable the default process-wide metrics in prom-client 9.0.0

leonerd:
Since prom-client v9, the default process-wide metrics are not actually enabled by default. (Despite what you'd imagine by the name ;) ). So we have to explicitly enable them.


lukebarnard1:
LGTM apart from maybe we should add `"prom-client":"^9.0.0"` to package.json `optionalDependencies`

leonerd:
PTAL - better plan is to only call the method if it exists, so we still work fine on older versions

PR #61, opened 2017-10-22
Provide an option for bridges to auto-group users

turt2live:
## Required PRs
* https://github.com/matrix-org/matrix-js-sdk/pull/570 - for the visibility component

## Overview
Many bridges may wish to group their users into a single community, making it abundantly clear in their portal rooms that the users are from a different network. A single flag (`groupLocalpart`) has been exposed as part of the `Bridge` constructor to enable this functionality. Bridges may also provide `dontJoinGroup` or `dontAdvertiseGroup` for users and rooms (respectively) to bypass the groups functionality on that particular entity.

## Migrating a bridge to use this
This PR does not add in the functionality to automatically join any created users to the group, as each bridge is different in how it accomplishes storing known users. This task is left up to the individual bridge. Below is an example startup routine for bridges that wish to do the migration automatically; some assumptions have been made about imports, naming, and structure.
```js
var bridge = new Bridge({ groupLocalpart: "irc", /* other options */ });
bridge.run(port).then(() => {
    var botClient = bridge.getBot().getClient();
    var userIds = /* a list of user IDs to add to the group */
    var group = botClient.getGroup("+irc:" + homeserverDomain);
    var promise = Promise.resolve(group);
    if (!group) {
        promise = botClient.createGroup("+irc:" + homeserverDomain);
    }

    return promise.then(group => {
        var chain = Promise.resolve();
        userIds.map(user => chain = chain.then(() => {
            return botClient.inviteUserToGroup("+irc:" + homeserverDomain, user);
        }.then(() => {
            var userClient = bridge.getIntent(user).getClient();
            return userClient.acceptGroupInvite("+irc:" + homeserverDomain);
        })));
        return chain;
    });
}).then(() => {
    var botClient = bridge.getBot().getClient();
    var roomIds = /* a list of room IDs to advertise flair in */

    var chain = Promise.resolve();
    roomIds.map(roomId => chain = chain.then(() => botClient.sendStateEvent(roomId, "m.room.related_groups", {groups: ["+irc:" + homeserverDomain]}, "")));
    return chain;
});
```

### DIFF
--- a/lib/bridge.js
+++ b/lib/bridge.js
@@ -170,6 +170,7 @@ function Bridge(opts) {
     this._queue = new EventQueue(this.opts.queue, this._onConsume.bind(this));
     this._prevRequestPromise = Promise.resolve();
     this._metrics = null; // an optional PrometheusMetrics instance
+    this._groupExists = false; // assumption; we'll update this on our first call
 }
 
 /**
@@ -585,11 +586,17 @@ Bridge.prototype.getIntentFromLocalpart = function(localpart, request) {
 
 Bridge.prototype._createGroupIfNotExists = function() {
     var self = this;
+    if (self._groupExists) {
+        // Don't hit the homeserver unless we're unsure about the group's state
+        return;
+    }
     return self._botClient.getGroupSummary(self.groupId).error(function() {
         return self._botClient.createGroup({
             localpart: self.opts.groupLocalpart,
             profile: {}
         });
+    }).then(function() {
+        self._groupExists = true;
     });
 };
 
--- a/lib/bridge.js
+++ b/lib/bridge.js
@@ -593,9 +593,8 @@ Bridge.prototype._getOrCreateGroup = function() {
         }).then(function(createdGroupId) {
             return self._botClient.getGroup(createdGroupId);
         });
-    } else {
-        return Promise.resolve(group);
     }
+    return Promise.resolve(group);
 };
 
 /**
@@ -644,9 +643,8 @@ Bridge.prototype.provisionUser = function(matrixUser, provisionedUser) {
                         // Advertise flair for the user by default
                         return newUser.setGroupPublicity(group.groupId, true);
                     });
-            } else {
-                throw new Error("Failed to get or create group " + self.groupId);
             }
+            throw new Error("Failed to get or create group " + self.groupId);
         });
     }
     return promise;
--- a/lib/bridge.js
+++ b/lib/bridge.js
@@ -88,6 +88,10 @@ var INTENT_CULL_EVICT_AFTER_MS = 1000 * 60 * 15; // 15 minutes
  * parameters in {@link Bridge~onEvent}. Disabling the context makes the
  * bridge do fewer database lookups, but prevents there from being a
  * <code>context</code> parameter. Default: false.
+ * @param {string=} opts.groupLocalpart If supplied, users and rooms will be associated
+ * with the group. If the group does not exist, a default one will be created prior to
+ * use. Users and rooms will not be migrated automatically - that is a task left up to
+ * the individual bridge.
  */
 function Bridge(opts) {
     if (typeof opts !== "object") {
@@ -131,6 +135,11 @@ function Bridge(opts) {
     }
     if (opts.disableContext === undefined) {
         opts.disableContext = false;
+    }
+
+    // define a group ID we can use as a reference
+    if (opts.groupLocalpart) {
+        this.groupId = "+" + opts.groupLocalpart + ":" + opts.domain;
     }
 
     // we'll init these at runtime
@@ -574,6 +583,16 @@ Bridge.prototype.getIntentFromLocalpart = function(localpart, request) {
     );
 };
 
+Bridge.prototype._getOrCreateGroup = function() {
+    var self = this;
+    var group = self._botClient.getGroup(self.groupId);
+    if (!group) {
+        return self._botClient.createGroup({localpart: self.opts.groupLocalpart, profile: {}}).then(function(createdGroupId) {
+            return self._botClient.getGroup(createdGroupId);
+        });
+    } else return Promise.resolve(group);
+};
+
 /**
  * Provision a user on the homeserver.
  * @param {MatrixUser} matrixUser The virtual user to be provisioned.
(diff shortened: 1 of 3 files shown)


lib/bridge.js
```diff
@@ -88,6 +88,10 @@ var INTENT_CULL_EVICT_AFTER_MS = 1000 * 60 * 15; // 15 minutes
  * parameters in {@link Bridge~onEvent}. Disabling the context makes the
  * bridge do fewer database lookups, but prevents there from being a
  * <code>context</code> parameter. Default: false.
+ * @param {string=} opts.groupLocalpart If supplied, users and rooms will be associated
+ * with the group. If the group does not exist, a default one will be created prior to
```
lukebarnard1: Worth mentioning at which point the users and rooms will be associated, I think. Also worth mentioning that they won't be unassociated and that users are invited whereas rooms are just associated.

lib/bridge.js
```diff
@@ -607,6 +633,19 @@ Bridge.prototype.provisionUser = function(matrixUser, provisionedUser) {
             return newUser.setAvatarUrl(provisionedUser.url);
         });
     }
+    if (!provisionedUser.dontJoinGroup && self.groupId) {
```
lukebarnard1: I think it'd be better to use a positively-named boolean (something like shouldJoinGroup) that defaults to true.

lib/bridge.js
```diff
@@ -651,6 +692,15 @@ Bridge.prototype._onAliasQuery = function(alias) {
             }
             // store the matrix room only
             return self._roomStore.setMatrixRoom(matrixRoom);
+        }).then(function () {
+            // Process group-related functions, if we need to.
+            if (!dontAdvertiseGroup && self.groupId) {
```
lukebarnard1: See previous comment, let's use a positively-named boolean

lib/bridge.js
```diff
@@ -651,6 +692,15 @@ Bridge.prototype._onAliasQuery = function(alias) {
...
+                return self._createGroupIfNotExists().then(function() {
+                    return self._botClient.sendStateEvent(roomId, "m.room.related_groups", {
```
lukebarnard1: This won't handle the case where there are other groups associated here, but I guess this is a non-issue for portal rooms?

lib/bridge.js
lukebarnard1: Actually, you might want to describe it totally differently and only in scope of the as-bridge library itself. AFAIC, the groupLocalpart causes three things to happen:
 - Creates the group when it is first required
 - Automatically joins ghost users to the group
 - Adds `m.room.related_rooms` event to rooms created via alias request with the group's ID in it

lib/bridge.js
turt2live: It shouldn't be an issue, although there's a chance that a bridge may shove some other related groups in before we get here. Adding in the safety wouldn't hurt.

lib/bridge.js
turt2live: This morning I was also thinking of changing it from a single string to an object, just to be able to provide a few more options. Namely for the future case of being able to control membership visibility. At this point it'd still just be the localpart though.

lukebarnard1: yep, sounds like something I should've picked up on, actually. WFM

turt2live:
@lukebarnard1 feedback incorporated. Thank you!

lib/bridge.js
```diff
@@ -702,10 +702,10 @@ Bridge.prototype._onAliasQuery = function(alias) {
             if (shouldAdvertiseGroup && self.groupId) {
                 return self._createGroupIfNotExists().then(function() {
                     return self._botClient.getStateEvent(roomId, "m.room.related_groups", "");
-                }).then(function(currentState) {
+                }).then(function(relatedGroupsContent) {
                     var groups = [];
-                    if (currentState && currentState.getContent()) {
-                        groups = currentState.getContent().groups || [];
+                    if (relatedGroupsContent && relatedGroupsContent.groups) {
```
lukebarnard1: I think we also should validate that `groups` is actually an array, for the sake of being paranoid.

lib/bridge.js
```diff
@@ -636,7 +636,7 @@ Bridge.prototype.provisionUser = function(matrixUser, provisionedUser) {
             return newUser.setAvatarUrl(provisionedUser.url);
         });
     }
-    if (!provisionedUser.dontJoinGroup && self.groupId) {
+    if (provisionedUser.shouldJoinGroup !== false && self.groupId) {
```
lukebarnard1: Maybe have a comment to explain that this is effectively default `true`?

lukebarnard1: Actually, given that you've done it the explicit way for `shouldAdvertiseGroup`, I'd just do them both the same way for readability and consistency.

lib/bridge.js
```diff
@@ -138,8 +141,8 @@ function Bridge(opts) {
     }
 
     // define a group ID we can use as a reference
-    if (opts.groupLocalpart) {
-        this.groupId = "+" + opts.groupLocalpart + ":" + opts.domain;
+    if (opts.group) {
```
lukebarnard1: Also need to null-guard `group.localpart`

turt2live:
@lukebarnard1 @dbkr This has been updated.

turt2live:
Blocking this on https://github.com/matrix-org/matrix-js-sdk/pull/570 to be able to support the visibility at the bridge level as well. Would probably be something like the following to declare the parts of a group:

```javascript
group: {
    localpart: "discord",
    memberVisibility: "private", // default "public"
}
```

turt2live:
The member visibility stuff is now implemented. It requires https://github.com/matrix-org/matrix-js-sdk/pull/570

lib/bridge.js
```diff
@@ -642,13 +651,20 @@ Bridge.prototype.provisionUser = function(matrixUser, provisionedUser) {
         shouldJoinGroup = false;
     }
 
+    var groupVisibility = self.opts.group.memberVisibility;
```
lukebarnard1: This should really only be in the scope of `if (shouldJoinGroup && self.groupId)

turt2live: moved

turt2live:
@lukebarnard1 linted, hopefully.

turt2live:
This also needs the js-sdk release, as the version bumped in this PR doesn't include the options change.

lukebarnard1:
right..

lukebarnard1:
FTR that release will be 0.9.1

turt2live:
Closing: This is being handled elsewhere in the bridge stuff.

lukebarnard1:
FTR I don't think the use of this has been precluded by https://github.com/matrix-org/synapse/pull/2686.

turt2live:
The bridge library would be better off having the registration modified to make https://github.com/matrix-org/synapse/pull/2686 happy. That's in a different area of the project from this, which to me is a new PR.

lukebarnard1:
OK, fair enough.

PR #67, opened 2018-06-18
Specify type=m.login.application_service when calling /register

Half-Shot:
This conforms to the ~~https://matrix.org/docs/spec/application_service/unstable.html#id14~~ https://matrix.org/docs/spec/application_service/unstable.html#server-admin-style-permissions spec item, and makes https://github.com/matrix-org/matrix-doc/issues/1312 less of an issue


turt2live:
id14 is https://matrix.org/docs/spec/application_service/unstable.html#server-admin-style-permissions

Half-Shot:
@turt2live Thanks, this never fails to bite me.

PR #70, opened 2018-07-23
Expose membership cache functions and use joined_members to cache

Half-Shot:
This PR exposes setMemberEntry and getMemberEntry to allow bridge writers greater control over the membership cache. The intention is you can save yourself some join or leave calls if you know the user has been joined. 

Also, the members returned from joined_members are added to the cache if the flag is set so that we don't rejoin users we already know are joined. We also cache these users as registered so the intent does not try to register them.

Finally, ``getMemberLists`` has been purposely broken due to the fact that /sync no longer works. Please shout if this should go in a different PR.

*This PR also makes eslint aware of es6 formatting, so the bloody thing will pass*


lib/components/app-service-bot.js
```diff
@@ -50,48 +56,60 @@ AppServiceBot.prototype.getJoinedRooms = function() {
...
+    //     var rooms = res.rooms || {};
+    //     var roomIdToRoom = rooms.join || {};
+    //     var dict = {};
+    //     Object.keys(roomIdToRoom).forEach(function(roomId) {
+    //         dict[roomId] = self._getRoomInfo(roomId, roomIdToRoom[roomId]);
+    //     })
+    //     return dict;
+    // });
+    throw new Error(
+        "The ASBot can no longer /sync because this functionality was removed." +
```
turt2live: s/ASBot/appservice bot/ please. Spelling things out is better than acronyms in my books.

lib/components/app-service-bot.js
```diff
@@ -50,48 +56,37 @@ AppServiceBot.prototype.getJoinedRooms = function() {
  * Get a map of joined user IDs for the given room ID. The values in the map are objects
  * with a 'display_name' and 'avatar_url' properties. These properties may be null.
  * @param {string} roomId The room to get a list of joined user IDs in.
+ * @param {boolean=false} storeMembership Store membership found in this call in the bridge cache.
```
turt2live: tbh I'm having troubles justifying tagging on a caching behaviour to these functions, which is more what I was trying to get at with splitting out the caching layer (sorry). What's the use case for this?

Half-Shot: Probably the most obvious thing is that this is returning the know joined state for users in a room, so it would be smart to store it in the cache that is used to get the joined state for users in a room.

Otherwise we end up where we are today with intents pointlessly joining users to rooms.

Half-Shot: We could also ask the bridges to implement this themselves, but that seems kinda silly since caching is kinda internal. And a friendly Boolean is somewhat nicer than asking developers to loop round the result and store it in a cache object they now need to keep track of.

turt2live: Can we not just use the cache they now (optionally) provide? I don't think we need the flag unless there's ever a case where you'd request membership and not want it to be fast.

Half-Shot: I'm failing to see how this would be slow?

lib/components/app-service-bot.js
```diff
@@ -5,11 +5,17 @@
  * @constructor
  * @param {MatrixClient} client The client instance configured for the AS bot.
  * @param {AppServiceRegistration} registration The registration that the bot
+ * @param {Function=} setJoined Called when a /joined_members request has a remote user.
```
anoadragon453: You put this in the middle of the previous line's explanation :p

Half-Shot: Actually, it was already like that :(

lib/components/app-service-bot.js
turt2live: By nature, not caching is slower than caching. So is there ever a use case where you wouldn't want to cache?

lib/components/app-service-bot.js
```diff
@@ -50,48 +56,34 @@ AppServiceBot.prototype.getJoinedRooms = function() {
...
+            }
+        }
         return res.joined;
     });
 };
 
 /**
- * Get the membership lists of every room the AS bot is in. This does a /sync.
- * @return {Promise} Resolves to a dict of room_id: {@link AppServiceBot~RoomInfo}.
+ * @throws {Error} This will always throw because /sync is no longer supported.
```
turt2live: minor thing, can we also throw a `@deprecated` on this?

lib/components/app-service-bot.js
```diff
@@ -6,10 +6,16 @@
  * @param {MatrixClient} client The client instance configured for the AS bot.
  * @param {AppServiceRegistration} registration The registration that the bot
  * is following. Used to determine which user IDs it is controlling.
+ * @param {Function=} setJoined Called when a /joined_members request has a remote user.
```
turt2live: In the future we might want to consider another controller class. Having the language generic would be a good thing to in the event that /joined_members gets replaced. The arguments to said function also need documenting.

And finally, what's the use case for this?

lib/components/app-service-bot.js
```diff
@@ -1,21 +1,17 @@
 "use strict";
-
```
turt2live: *aneurysm*

Half-Shot:
YAY!

PR #73, opened 2018-07-24
Don't resolve _ensureRegistered if we cannot ensure a user is registered.

### LINKED ISSUE #71: _ensureRegistered() will always be resolved, even when failed.

Half-Shot:
I think that's fairly sane?

Fixes #71

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,7 +1,8 @@
 {
   "env": {
     "node": true,
-    "jasmine": true
+    "jasmine": true,
+    "es6": true,
   },
   "rules": {
     "consistent-return": 2,
--- a/lib/components/intent.js
+++ b/lib/components/intent.js
@@ -649,17 +649,17 @@ Intent.prototype._ensureRegistered = function() {
     if (this.opts.registered) {
         return Promise.resolve("registered=true");
     }
-    var self = this;
-    var userId = self.client.credentials.userId;
-    var localpart = new MatrixUser(userId).localpart;
-    return self.botClient.register(localpart).then(function(res) {
-        self.opts.registered = true;
+    const userId = this.client.credentials.userId;
+    const localpart = new MatrixUser(userId).localpart;
+    return self.botClient.register(localpart).then((res) => {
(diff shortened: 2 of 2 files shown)


.eslintrc
```diff
@@ -1,7 +1,8 @@
 {
   "env": {
     "node": true,
-    "jasmine": true
+    "jasmine": true,
+    "es6": true,
```
turt2live: I think this is being dealt with elsewhere?

lib/components/intent.js
```diff
@@ -649,17 +649,17 @@ Intent.prototype._ensureRegistered = function() {
...
+    const localpart = new MatrixUser(userId).localpart;
+    return self.botClient.register(localpart).then((res) => {
+        this.opts.registered = true;
         return res;
-    }, function(err) {
+    }).catch((err) => {
         if (err.errcode === "M_USER_IN_USE") {
-            self.opts.registered = true;
+            this.opts.registered = true;
             return err;
```
turt2live: If we're rejecting the promise, we should probably suppress this error or let it fall through to the throw below, whichever is best.

Half-Shot: Supression works for me.

.eslintrc
Half-Shot: This was to stop it from bitching about consts and arrow funcs, sorry. We've already merged this change in, so it should be fine.

lib/components/intent.js
```diff
@@ -649,17 +649,17 @@ Intent.prototype._ensureRegistered = function() {
...
+    return self.botClient.register(localpart).then((res) => {
+        this.opts.registered = true;
         return res;
-    }, function(err) {
+    }).catch((err) => {
         if (err.errcode === "M_USER_IN_USE") {
-            self.opts.registered = true;
-            return err;
+            this.opts.registered = true;
+            return {};
```
turt2live: I think you can return null and get the same effect. Objects feel a bit weird to me, and would appreciate a comment if kept.

PR #75, opened 2018-07-25
Add Intent.setPresence

Half-Shot:
Also adds a ``Intent`` opt ``enablePresence`` which is true by default. Useful for bridges who want to disable presence support without having to go through all the presence code.

### DIFF
--- a/lib/components/intent.js
+++ b/lib/components/intent.js
@@ -48,11 +48,15 @@ var STATE_EVENT_TYPES = [
  * @param {boolean} opts.dontJoin True to not attempt to join a room before
  * sending messages into it. The surrounding code will have to ensure the correct
  * membership state itself in this case. Default: false.
+ * 
+ * @param {boolean} [opts.enablePresence=true] True to send presence, false to no-op.
  */
 function Intent(client, botClient, opts) {
     this.client = client;
     this.botClient = botClient;
     opts = opts || {};
+
+    opts.enablePresence = !(opts.enablePresence !== false);
 
     if (opts.backingStore) {
         if (!opts.backingStore.setPowerLevelContent ||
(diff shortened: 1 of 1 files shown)


lib/components/intent.js
```diff
@@ -48,12 +48,16 @@ var STATE_EVENT_TYPES = [
...
  * membership state itself in this case. Default: false.
+ * 
+ * @param {boolean} [opts.enablePresence=true] True to send presence, false to no-op.
  */
 function Intent(client, botClient, opts) {
     this.client = client;
     this.botClient = botClient;
     opts = opts || {};
 
+    opts.enablePresence = !(opts.enablePresence !== false);
```
turt2live: or `!!opts.enablePresence`? I assume you're trying to achieve a default true unless false here, which a double negative can do.

lib/components/intent.js
```diff
@@ -470,6 +474,22 @@ Intent.prototype.createAlias = function(alias, roomId) {
     });
 };
 
+/**
+ * Set the presence of this user.
+ * @param {string} presence One of "online", "offline" or "unavailable".
+ * @param {string} status_msg The status message to attach.
+ * @return {Promise}
+ */
+Intent.prototype.setPresence = function(presence, status=undefined) {
```
turt2live: `status` vs `status_msg` in the docs

lib/components/intent.js
```diff
@@ -470,6 +474,22 @@ Intent.prototype.createAlias = function(alias, roomId) {
     });
 };
 
+/**
+ * Set the presence of this user.
+ * @param {string} presence One of "online", "offline" or "unavailable".
+ * @param {string} status_msg The status message to attach.
+ * @return {Promise}
```
turt2live: What does the promise do?

lib/components/intent.js
Half-Shot: Feelin more dumb now :). I'll update.

lib/components/intent.js
Half-Shot: http://matrix-org.github.io/matrix-js-sdk/0.10.3/module-client-MatrixClient.html#setPresence Not sure which I should be following

turt2live: `Math.random() <= 0.5 ? "status" : "status_msg"` (aka: pick one :p )

PR #83, opened 2018-08-23
Request caching for intents

Half-Shot:
Following up from https://github.com/matrix-org/matrix-appservice-irc/pull/658, this is a more generic caching system to allow intents to cache simple requests and give them a TTL. The idea being that bridges won't have to create their own specialised caches. This is entirely configurable.

The changes are:
- ``getProfileInfo`` caches by default, anticipating that bridges won't care if a profile is up to 30 seconds behind.
- ``getEvent`` caches by default, since events are unlikely to change. There is a possibility that events will no longer be accessible to the client, but we hope people aren't using this to determine access to a room.
- ``roomState`` does not cache by default, because this call can be quite important to how rooms get bridged. However, the option is there.
- Intent now takes a caching.ttl option to set the time to live for all caches, which will default to 30.

Open questions:

- Do we want per cache ttls/sizes to be a thing?
- Do we want to enable all/disable all caches by default?


lib/bridge.js
```diff
@@ -17,7 +17,7 @@ const yaml = require("js-yaml");
 const Promise = require("bluebird");
 const Datastore = require("nedb");
 const util = require("util");
-const MembershipCache = new require("./components/membership-cache");
+const MembershipCache = require("./components/membership-cache");
```
turt2live: js never ceases to amaze me

lib/components/client-request-cache.js
```diff
@@ -0,0 +1,72 @@
+/**
+ * Caches requests in memory and handles expiring them.
+ */
+class ClientRequestCache {
+    /**
+     * @param ttl {Number} How old a result can be before it gets expired.
+     * @param size {Number} How many results to store before we trim.
+     * @param requestFunc The function to use when requesting.
```
turt2live: on cache miss, I assume?

lib/components/client-request-cache.js
```diff
@@ -0,0 +1,72 @@
...
+            throw Error("'requestFunc' must be a function");
+        }
+        this._requestContent = new Map(); // key => {ts, content}
+        this.requestFunc = requestFunc;
+        this.ttl = ttl;
+        this.maxSize = size;
+    }
+
+    /**
+     * Get's a result of a request from the cache, or otherwise
```
turt2live: > Get's

lib/components/client-request-cache.js
```diff
@@ -0,0 +1,72 @@
...
+                    content: result,
+                });
+                if (this._requestContent.size > this.maxSize) {
+                    const oldKey = this._requestContent.keys().next().value;
+                    this._requestContent.delete(oldKey);
+                }
+            }
+            return result;
+        }).catch((err) => {
+            // If the request failed for any reason we want to delete it.
```
turt2live: we don't delete it though?

lib/components/displayname-cache.js
```diff
@@ -0,0 +1,54 @@
+/**
+ * Caches requests in memory and handles expiring them.
+ */
+class RequestCache {
```
turt2live: This file doesn't have a displayname cache, it has a copy/paste of the request cache.

lib/components/intent.js
```diff
@@ -1,13 +1,16 @@
...
+const MatrixEvent = require("matrix-js-sdk").MatrixEvent;
+const RoomMember = require("matrix-js-sdk").RoomMember;
+const ClientRequestCache = require("./client-request-cache");
 
-var STATE_EVENT_TYPES = [
+const STATE_EVENT_TYPES = [
     "m.room.name", "m.room.topic", "m.room.power_levels", "m.room.member",
     "m.room.join_rules", "m.room.history_visibility"
 ];
+const DEFAULT_CACHE_TTL = 30000;
```
turt2live: I think we can safely up this to a minute or two.

lib/components/intent.js
```diff
@@ -97,6 +103,34 @@ function Intent(client, botClient, opts) {
         }
     }
 
+    if (!opts.caching) {
+        opts.caching = {
+
```
turt2live: does the linter require this much whitespace?

lib/components/intent.js
```diff
@@ -97,6 +103,34 @@ function Intent(client, botClient, opts) {
...
+    }
+
+    opts.caching.ttl = opts.caching.ttl === undefined ? DEFAULT_CACHE_TTL : opts.caching.ttl;
+    opts.caching.size = opts.caching.size === undefined ? DEFAULT_CACHE_SIZE : opts.caching.ttl;
+    this._requestCaches = {};
+    this._requestCaches.profile = new ClientRequestCache(
+        opts.caching.ttl,
+        opts.caching.size,
+        (_, userId, info) => {
+        return this.intent.getProfileInfo(userId, info, false);
```
turt2live: indentation (on this and the 2 below).

I'd also recommend having the closing curly not be part of the closing parens

spec/unit/client-request-cache.spec.js
```diff
@@ -0,0 +1,120 @@
...
+const Promise = require("bluebird");
+
+describe("ClientRequestCache", function() {
+    describe("constructor", function() {
+        it("Can construct", function() {
+            const crc = new ClientRequestCache(50, 1, () => { });
+            expect(crc.ttl).toBe(50);
+            expect(crc.maxSize).toBe(1);
+            expect(crc.requestFunc).toBeDefined();
+            expect(crc._requestContent).toBeDefined();
```
turt2live: Checking internal properties feels wrong

spec/unit/client-request-cache.spec.js
```diff
@@ -0,0 +1,120 @@
...
+            return crc.get("1").then((res) => {
+                expect(res).toBe("Thing 1!");
+                return crc.get("2");
+            }).then((res) => {
+                expect(res).toBe("Thing 2!");
+            });
+        });
+        it("should pass down failures", () => {
+            const crc = new ClientRequestCache(1000, 1, () => {
+                return Promise.reject("Sorry, this test has subject to a GDPR request.");
```
turt2live: can we also have another test for if the function throws?

spec/unit/client-request-cache.spec.js
```diff
@@ -0,0 +1,120 @@
...
+        });
+        it("should respect max size", () => {
+            const crc = new ClientRequestCache(1000, 2, (key) => `${key}baa`);
+            return Promise.all([
+                crc.get("1"),
+                crc.get("2"),
+                crc.get("3"),
+                crc.get("4"),
+            ]).then(() => {
+                expect([...crc._requestContent.values()].map((v) => v.content)).toEqual([
```
turt2live: A public `getAllItems` or similar function would be preferred instead of a poke at internal properties.

lib/bridge.js
Half-Shot: Sorry, it had to be done.

lib/components/displayname-cache.js
Half-Shot: This one was meant to be deleted :/

lib/components/intent.js
Half-Shot: I think 90 seconds is a pretty standard one?

spec/unit/client-request-cache.spec.js
Half-Shot: We can probably assume they got defined if it didn't throw I guess.

PR #85, opened 2018-08-29
Room linking validation

Half-Shot:
This change adds a new class, the RoomLinkValidator. This is a rather bruteforced method of avoiding double bridged rooms by creating a rule file and setting rules about what can and can't be bridged by checking what userIds are present in the room.

~~The rule file can optionally be watched for changes, and valid changes will be applied on the fly.~~

The rule file can be reloaded by hitting the ``/_bridge/roomLinkValidator/reload`` endpoint.

### DIFF
--- a/lib/bridge.js
+++ b/lib/bridge.js
@@ -760,9 +760,6 @@ Bridge.prototype._onEvent = function(event) {
 Bridge.prototype._onConsume = function(err, data) {
     if (!err) {
         this.opts.controller.onEvent(data.request, data.context);
-        if (this._roomLinkValidator) {
-            this._roomLinkValidator.onEvent(data.request, data.content);
-        }
         return;
     }
     if (!this.opts.controller.onLog) {
--- a/lib/bridge.js
+++ b/lib/bridge.js
@@ -17,7 +17,8 @@ const yaml = require("js-yaml");
 const Promise = require("bluebird");
 const Datastore = require("nedb");
 const util = require("util");
-const MembershipCache = new require("./components/membership-cache");
+const MembershipCache = require("./components/membership-cache");
+const RoomLinkValidator = require("./components/room-link-validator").RoomLinkValidator;
 
 // The frequency at which we will check the list of accumulated Intent objects.
 const INTENT_CULL_CHECK_PERIOD_MS = 1000 * 60; // once per minute
@@ -163,6 +164,7 @@ function Bridge(opts) {
     this._queue = new EventQueue(this.opts.queue, this._onConsume.bind(this));
     this._prevRequestPromise = Promise.resolve();
     this._metrics = null; // an optional PrometheusMetrics instance
+    this._roomLinkValidator = null;
 }
 
 /**
@@ -231,6 +233,14 @@ Bridge.prototype.run = function(port, config, appServiceInstance) {
     this._appServiceBot = new AppServiceBot(
         this._botClient, self.opts.registration, this._membershipCache
     );
+
+    if (this.opts.roomLinkingRules !== undefined) {
+        this._roomLinkValidator = new RoomLinkValidator(
+            this.opts.roomLinkingRules,
+            this._appServiceBot
+        );
+    }
+
     this._requestFactory = new RequestFactory();
     if (this.opts.controller.onLog && this.opts.logRequestOutcome) {
         this._requestFactory.addDefaultResolveCallback(function(req, res) {
@@ -532,6 +542,10 @@ Bridge.prototype.getBot = function() {
     return this._appServiceBot;
 };
 
+Bridge.prototype.getRoomLinkValidator = function() {
+    return this._roomLinkValidator;
+}
+
 /**
  * Retrieve an Intent instance for the specified user ID. If no ID is given, an
  * instance for the bot itself is returned.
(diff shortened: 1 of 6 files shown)


lib/components/room-link-validator.js
```diff
@@ -0,0 +1,168 @@
...
+            this.config = new ConfigValidator(RULE_SCHEMA);
+            this.readRuleFile(config.ruleFile);
+            if (config.watch) {
+                this.log(`Watching for rule changes on ${config.ruleFile}`);
+                this.config.watchForChanges(config.ruleFile, () => {
+                    try {
+                        this.readRuleFile(config.ruleFile);
+                    }
+                    catch (e) {
+                        // Catch these because we don't want to blow up in here.
```
turt2live: we should at least log the errors

lib/components/room-link-validator.js
```diff
@@ -0,0 +1,168 @@
...
+        else if (config.rules) {
+            this.rules = this.evaulateRules(config.rules);
+        }
+        else {
+            throw new Error("Either config.ruleFile or config.rules must be set");
+        }
+    }
+
+    readRuleFile(filename) {
+        try {
```
turt2live: I think it's better to have the catch high up (where my comment about logging things is) and just have this throw.

lib/components/room-link-validator.js
```diff
@@ -0,0 +1,168 @@
...
+            this.log(`Detected rule config change..`);
+            const rules = this.config.validate(filename);
+            if (rules === undefined) {
+                throw Error("Rule file contents was undefined");
+            }
+            this.log(`Rule file ok, checking rules..`);
```
turt2live: OCD: 3 full stops

lib/components/room-link-validator.js
```diff
@@ -0,0 +1,168 @@
...
+        newRules.userIds.conflict = rules.userIds.conflict.map(
+            (regexStr) => new RegExp(regexStr)
+        );
+        newRules.userIds.exempt = rules.userIds.exempt.map(
+            (regexStr) => new RegExp(regexStr)
+        );
+        return newRules;
+    }
+
+    validateRoom (roomId, cache=true) {
```
turt2live: the other functions don't have a space before `(`.

lib/bridge.js
```diff
@@ -717,6 +758,9 @@ Bridge.prototype._onEvent = function(event) {
 Bridge.prototype._onConsume = function(err, data) {
     if (!err) {
         this.opts.controller.onEvent(data.request, data.context);
+        if (this._roomLinkValidator) {
+            this._roomLinkValidator.onEvent(data.request, data.content);
```
turt2live: The validator doesn't have an `onEvent` though?

turt2live:
@Half-Shot any updates on this?

Half-Shot:
Going through it now.

lib/components/room-link-validator.js
Half-Shot: And now we can, thanks to logging!

turt2live:
@Half-Shot you have build failures

lib/bridge.js
```diff
@@ -291,6 +312,23 @@ Bridge.prototype._customiseAppservice = function() {
     if (this.opts.controller.thirdPartyLookup) {
         this._customiseAppserviceThirdPartyLookup(this.opts.controller.thirdPartyLookup);
     }
+    if (this.opts.roomLinkValidation && this.opts.roomLinkValidation.triggerEndpoint) {
+        this.addAppServicePath({
+            method: "POST",
+            path: "/_bridge/roomLinkValidator/reload",
```
turt2live: I don't think this adds auth?

lib/bridge.js
```diff
@@ -291,6 +312,23 @@ Bridge.prototype._customiseAppservice = function() {
...
+            handler: function(req, res) {
+                try {
+                    // Will use filename if provided, or the config
+                    // one otherwised.
+                    this._roomLinkValidator.readRuleFile(req.query.filename);
+                    res.status(200).send("Success");
+                } catch (e) {
+                    res.status(500).send("Failed: " + e);
+                }
+
```
turt2live: whitespace

lib/bridge.js
```diff
@@ -534,6 +572,25 @@ Bridge.prototype.getBot = function() {
     return this._appServiceBot;
 };
 
+/**
+ * Can the room be provisioned based on the rules provided to the
+ * bridge.
```
turt2live: questions end with question marks. Alternatively, it could be rephrased as "Whether or not the room can be..."

lib/components/config-validator.js
```diff
@@ -47,6 +47,7 @@ ConfigValidator.prototype.validate = function(inputConfig, defaultConfig) {
     return extend(true, defaultConfig, inputConfig);
 }
 
+
```
turt2live: whitespace?

lib/components/logging.js
```diff
@@ -191,24 +191,24 @@ if (winston) {
...
 }
 
 module.exports = {
     get: (name) => {
-        return instance.get(name);
+        return instance.Get(name);
     },
     configure: (config) => {
-        instance.configure(config);
+        instance.Configure(config);
```
turt2live: these both should be lowercase though :( Why was the build complaining about it?

lib/components/logging.js
```diff
@@ -191,13 +191,13 @@ if (winston) {
...
     instance = {
-        get: (name) => {
+        Get: (name) => {
             const logWrapper = new LogWrapper();
             // Console has all the functions already.
             logWrapper.setLogger(console);
             return logWrapper;
         },
-        configure: () => {
+        Configure: () => {
```
turt2live: gah, my comment went missing. These should be lowerCamelCase please

lib/bridge.js
Half-Shot: it does now :)

PR #90, opened 2018-10-01
Add Logging component

Half-Shot:
This one is based off gitter's logger. It optionally depends on winston but will use the console if winston is not available.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,41 @@ The component which orchestrates other components: a "glue" component. Provides
 a way to start the bridge. This is the component most examples use. Has
 dependencies on most of the components listed above.
 
+## `Logging`
+This component exposes access to the bridges log reporter. To use, you should
+install the optional packages `winston`, `winston-daily-rotate-file` and
+`chalk` to get nice formatted log lines, otherwise it will default to the JS
+console. To use the component, use `Logging.Configure(configObject)` to setup
+the logger, which takes the following options:
+```
+    # A level to set the console reporter to.
+    console: "error|warn|info|debug|off"
+
+    # Format to append to log files.
+    fileDatePattern: "YYYY-MM-DD",
+
+    # Format of the timestamp in log files.
+    timestampFormat: "MMM-D HH:mm:ss.SSS"
+
+    # Log files to emit to, keyed of the minimum level they report.
+    files: {
+        "abc.log" => "error|warn|info|debug|off"
+    }
+
+    # The maximum number of files per level before old files get cleaned
+    up. Use 0 to disable.
+    maxFiles: 5
+```
+
+You **MUST** configure the logger before anything will be emitted to the console.
+
+You can then use `Logging.Get(ModuleName)` to start logging to the reporter,
+using the `error, warn, info, debug` functions. Arguments to these functions will
+automatically be seralized if they aren't strings.
+
+NOTE: ``opts.controller.onLog`` will override this, but if not set then the logging
+transport is used.
+
 ## Data Models
  * `MatrixRoom` - A representation of a matrix room.
  * `RemoteRoom` - A representation of a third-party room.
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -94,6 +94,41 @@ The component which orchestrates other components: a "glue" component. Provides
 a way to start the bridge. This is the component most examples use. Has
 dependencies on most of the components listed above.
 
+## `Logging`
+This component exposes access to the bridges log reporter. To use, you should
+install the optional packages `winston`, `winston-daily-rotate-file` and
+`chalk` to get nice formatted log lines, otherwise it will default to the JS
```
turt2live: we might want to recommend versions here, like `winston@3` or whatever

README.md
```diff
@@ -94,6 +94,41 @@ The component which orchestrates other components: a "glue" component. Provides
...
+console. To use the component, use `Logging.Configure(configObject)` to setup
+the logger, which takes the following options:
+```
+    # A level to set the console reporter to.
+    console: "error|warn|info|debug|off"
+
+    # Format to append to log files.
+    fileDatePattern: "YYYY-MM-DD",
```
turt2live: the others don't have a comma, and they probably should. Can we get away with putting the surrounding curly braces around the whole snippet and syntax highlight it for javascript? This is purely aesthetic and nothing else

README.md
```diff
@@ -94,6 +94,41 @@ The component which orchestrates other components: a "glue" component. Provides
...
+
+    # The maximum number of files per level before old files get cleaned
+    up. Use 0 to disable.
+    maxFiles: 5
+```
+
+You **MUST** configure the logger before anything will be emitted to the console.
+
+You can then use `Logging.Get(ModuleName)` to start logging to the reporter,
+using the `error, warn, info, debug` functions. Arguments to these functions will
```
turt2live: nitpick: `error`, `warn`, etc (don't include the comma in the code tag)

lib/bridge.js
```diff
@@ -18,6 +18,7 @@ const Promise = require("bluebird");
 const Datastore = require("nedb");
 const util = require("util");
 const MembershipCache = require("./components/membership-cache");
+const log = require("./components/logging").Get("Bridge");
```
turt2live: general principle: the log name should match the file name including casing (`bridge`, `cli`, etc)

README.md
```diff
@@ -94,6 +94,41 @@ The component which orchestrates other components: a "glue" component. Provides
...
+    # A level to set the console reporter to.
+    console: "error|warn|info|debug|off"
+
+    # Format to append to log files.
+    fileDatePattern: "YYYY-MM-DD",
+
+    # Format of the timestamp in log files.
+    timestampFormat: "MMM-D HH:mm:ss.SSS"
+
+    # Log files to emit to, keyed of the minimum level they report.
```
turt2live: Should also say that not including the option (or having it falsey) disables files

README.md
```diff
@@ -94,6 +94,41 @@ The component which orchestrates other components: a "glue" component. Provides
...
+    files: {
+        "abc.log" => "error|warn|info|debug|off"
```
turt2live: it's actually `"abc.log": "error|warn|etc"`?

README.md
Half-Shot: I kinda hoped it would be implied by the package.json, as I think you can ask npm to install optionals.

turt2live: I tried to find information on npm installing deps that nested in a project, but couldn't find anything. Duplicating the versions (even just the major versions) is fine here, I think.

README.md
Half-Shot: Or "/tmp/abc.log", it can be relative.

turt2live: I mean more that the arrow is not the right syntax, is it?

PR #91, opened 2018-10-02
Bump packages and replace jayschema

Half-Shot:
This is awaiting me doing a release for matrix-appservice-node, but anyway this resolves a lot of our security issues.

### DIFF
--- a/lib/components/config-validator.js
+++ b/lib/components/config-validator.js
@@ -1,7 +1,7 @@
 "use strict";
 const fs = require("fs");
 const yaml = require("js-yaml");
-const validator = require('is-my-json-valid');
+const Validator = require("is-my-json-valid");
 const extend = require("extend");
 
 /**
@@ -29,7 +29,7 @@ ConfigValidator.prototype.validate = function(inputConfig, defaultConfig) {
     if (typeof inputConfig === "string") {
         inputConfig = this._loadFromFile(inputConfig);
     }
-    const js = new validator(this.schema, {
+    const js = Validator(this.schema, {
         verbose: true,
     });
     const res = js(inputConfig, this.schema);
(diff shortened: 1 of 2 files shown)


lib/components/config-validator.js
```diff
@@ -1,8 +1,8 @@
 "use strict";
-var fs = require("fs");
-var yaml = require("js-yaml");
-var JaySchema = require("jayschema");
-var extend = require("extend");
+const fs = require("fs");
+const yaml = require("js-yaml");
+const validator = require('is-my-json-valid');
```
turt2live: quotes are different

turt2live: also can we `UpperCamelCase` the validator given it's a class?

turt2live: but it's a class that gives a function... gah why is js so horrible sometimes :/ I still think upper camel case is the way to go here, even if the package is weird.

package.json
```diff
@@ -26,19 +26,20 @@
   "dependencies": {
     "bluebird": "^2.9.34",
     "extend": "^3.0.0",
-    "jayschema": "^0.3.1",
-    "js-yaml": "^3.4.0",
-    "matrix-appservice": "^0.3.4",
+    "is-my-json-valid": "^2.19.0",
+    "js-yaml": "^3.12.0",
+    "matrix-appservice": "matrix-org/matrix-appservice-node#73839c6077b92b4cd0d52ec8dee599a96142d3b3",
```
turt2live: I'd rather us wait for the package to be released before we merge this PR

lib/components/config-validator.js
Half-Shot: I was making it conform to the others, but yea

package.json
Half-Shot: Seems fair

PR #100, opened 2019-03-11
Support room upgrades

### LINKED ISSUE #95: Support room upgrades via tombstones
This would be best handled by watching for tombstone events and firing a handler to deal with it, as it will likely need manual intervention by the bridges themselves.

Half-Shot:
Fixes #95 

This works by listening for a `m.room.tombstone` event and attempts to join the referenced room (first by room_id, then by alias). If that succeeds it finds all entries by the old room id and replace with the new room id. If the join fails, it waits for a subsequent invite before trying to update the entries.

The entry update logic can be overriden by providing a callback, and bridges can hook into another callback once an upgrade has been completed.

This functionality is disabled by default and must be enabled by providing a config.

### DIFF
--- a/lib/bridge.js
+++ b/lib/bridge.js
@@ -741,7 +741,7 @@ Bridge.prototype._onEvent = function(event) {
             this.opts.registration.isUserMatch(event.user_id, true)) {
         return Promise.resolve();
     }
-
+    // m.room.tombstone is the event that signals a room upgrade.
     if (event.type === "m.room.tombstone" && this._roomUpgradeHandler) {
         this._roomUpgradeHandler.onTombstone(ev);
         if (this.opts.RoomUpgradeHandler.consumeEvent) {
@@ -751,7 +751,10 @@ Bridge.prototype._onEvent = function(event) {
     else if (event.type === "m.room.member" &&
                event.state_key === this._appServiceBot.getUserId() &&
                event.content.membership === "invite") {
-        if (this._roomUpgradeHandler.onInvite(event) &&
+        // A invite-only room that has been upgraded won't have been joinable,
+        // so we are listening for any invites to the new room.
+        const isUpgradeInvite = this._roomUpgradeHandler.onInvite(event);
+        if (isUpgradeInvite &&
             this.opts.RoomUpgradeHandler.consumeEvent) {
             return Promise.resolve();
         }
--- a/lib/bridge.js
+++ b/lib/bridge.js
@@ -743,7 +743,7 @@ Bridge.prototype._onEvent = function(event) {
     }
 
     if (event.type === "m.room.tombstone" && this._roomUpgradeHandler) {
-        return this._roomUpgradeHandler.onTombstone(ev);
+        this._roomUpgradeHandler.onTombstone(ev);
         if (this.opts.RoomUpgradeHandler.consumeEvent) {
             return Promise.resolve();
         }
--- a/lib/bridge.js
+++ b/lib/bridge.js
@@ -184,8 +184,12 @@ function Bridge(opts) {
     this._prevRequestPromise = Promise.resolve();
     this._metrics = null; // an optional PrometheusMetrics instance
     this._roomLinkValidator = null;
-    this._roomUpgradeHandler = opts.roomUpgradeOpts ?
-        new RoomUpgradeHandler(opts.roomUpgradeOpts, this) : null;
+    if (opts.roomUpgradeOpts) {
+        this.opts.roomUpgradeOpts.consumeEvent = opts.roomUpgradeOpts.consumeEvent !== false ? true : false;
+        this._roomUpgradeHandler = new RoomUpgradeHandler(opts.roomUpgradeOpts, this);
+    } else {
+        this._roomUpgradeHandler = null;
+    }
 }
 
 /**
--- a/lib/bridge.js
+++ b/lib/bridge.js
@@ -20,6 +20,7 @@ const util = require("util");
 const MembershipCache = require("./components/membership-cache");
 const RoomLinkValidator = require("./components/room-link-validator").RoomLinkValidator;
 const RLVStatus = require("./components/room-link-validator").validationStatuses;
+const RoomUpgradeHandler = require("./components/room-upgrade-handler");
 
 const log = require("./components/logging").get("bridge");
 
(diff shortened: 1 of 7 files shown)


lib/bridge.js
```diff
@@ -733,7 +742,21 @@ Bridge.prototype._onEvent = function(event) {
...
+    if (event.type === "m.room.tombstone" && this._roomUpgradeHandler) {
+        this._roomUpgradeHandler.onTombstone(ev);
+        if (this.opts.RoomUpgradeHandler.consumeEvent) {
+            return Promise.resolve();
+        }
+    }
+    else if (event.type === "m.room.member" &&
+               event.state_key === this._appServiceBot.getUserId() &&
+               event.content.membership === "invite") {
+        if (this._roomUpgradeHandler.onInvite(event) &&
```
turt2live: Why are we calling an upgrade handler with invites?

(Hint: add a comment)

lib/components/intent.js
```diff
@@ -466,7 +466,7 @@ Intent.prototype.leave = function(roomId) {
  * @param {string} userId The ID of the user whose profile to return
  * @param {string} info The profile field name to retrieve (e.g. 'displayname'
  * or 'avatar_url'), or null to fetch the entire profile information.
- * @param {boolean} [useCache=true] Should the request attempt to lookup
+ * @param {boolean} [uuserIdseCache=true] Should the request attempt to lookup
```
turt2live: wat

lib/components/room-upgrade-handler.js
```diff
@@ -0,0 +1,135 @@
...
+        log.info(`Got tombstone event for ${ev.room_id} -> ${movingTo}`);
+        // Try to join the new room.
+        return this._joinNewRoom(movingTo, movingTo).then(() => {
+            return this._onJoinedNewRoom(ev.room_id, movingTo);
+        }).catch((err) => {
+            log.error("Couldn't handle room upgrade: ", err);
+            // We can wait for an invite and try again.
+            this._waitingForInvite.set(movingTo, ev.room_id);
+        });
+
```
turt2live: nit: blank line

lib/components/room-upgrade-handler.js
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    _joinNewRoom(newRoomId, roomIdOrAlias) {
+        const intent = this._bridge.getIntent();
+        return intent.join(roomIdOrAlias, true, true).catch((ex) => {
+            if (newRoomId !== roomIdOrAlias ||
+                !(ex.errcode == "M_UNKNONW" && ex.error === "No known servers")) {
+                // We need to wait to be invited
```
turt2live: You're joining wrong if you need an invite. The spec says to use the server that sent the tombstone event.

lib/components/room-upgrade-handler.js
```diff
@@ -0,0 +1,135 @@
...
+                throw Error("Need to wait for invite");
+            }
+            // RoomId is not routable, so try to get the alias and join that.
```
turt2live: This is unsafe and shouldn't be done. Throw an error instead and let the appservice deal with it

lib/components/room-upgrade-handler.js
```diff
@@ -0,0 +1,135 @@
+const log = require("./logging").get("RoomUpgradeHandler");
+const MatrixRoom = require("../models/rooms/matrix");
+
+/**
+ * Handles migration of rooms when a room upgrade is performed.
+ */
+class RoomUpgradeHandler {
+    /**
+     * @param {RoomUpgradeHandler~Options} opts
```
turt2live: what are they?

lib/bridge.js
```diff
@@ -105,6 +109,8 @@ const INTENT_CULL_EVICT_AFTER_MS = 1000 * 60 * 15; // 15 minutes
  * @param {boolean=} opts.roomLinkValidation.triggerEndpoint Enable the endpoint
  * to trigger a reload of the rules file.
  * Default: false
+ * @param {RoomUpgradeHandler~Options} opts.roomUpgradeOpts Options to supply to
+ * the room upgrade  handler. If not defined then no upgrades are performed.
```
turt2live: Upgrades happen regardless of whether or not this is a thing. Instead, can we say "If not defined, the bridge will not handle upgrades"?

turt2live:
also this needs options for handling ghosts.

lib/components/room-upgrade-handler.js
Half-Shot: Ah, gotcha

Half-Shot: Even in the case of invite-only rooms?

lib/components/room-upgrade-handler.js
Half-Shot: See below

Half-Shot:
>also this needs options for handling ghosts.

It doesn't. Ghosts are very particular to the bridge and should be handled on `onRoomMigrated`

lib/components/room-upgrade-handler.js
turt2live: They should be described here in the jsdoc if we expect people to create this object.

Half-Shot:
>This also needs proper documentation outside of the code.

More than JSDoc?

lib/bridge.js
```diff
@@ -732,8 +745,25 @@ Bridge.prototype._onEvent = function(event) {
...
         return Promise.resolve();
     }
+    // m.room.tombstone is the event that signals a room upgrade.
+    if (event.type === "m.room.tombstone" && this._roomUpgradeHandler) {
+        this._roomUpgradeHandler.onTombstone(ev);
+        if (this.opts.RoomUpgradeHandler.consumeEvent) {
+            return Promise.resolve();
+        }
+    }
+    else if (event.type === "m.room.member" &&
```
turt2live: pick a bracket style and stick with it: either `}\nelse` or `} else`. Applies to catches too.

lib/components/room-upgrade-handler.js
turt2live: that's fair, I thought this was much higher in the stack. Do we even need jsdoc here?

lib/components/room-upgrade-handler.js
turt2live: technically, yes. Looks like you've handled this correctly.

lib/bridge.js
Half-Shot: `}\nelse` is what apparently what `ESLint` wants.

lib/components/room-upgrade-handler.js
Half-Shot: Yes, otherwise we don't know the type that we are passing.

lib/bridge.js
Half-Shot: I'm failing to see where abouts in my changes that it's inconsistent..

turt2live: line 190 of this file vs here

Half-Shot: Well spotted

PR #101, opened 2019-03-13
Don't consider a user registered if they are banned or invited

### LINKED ISSUE #96: Membership cache treats "invited" as a join and thus believes the user exists
Which isn't true, you can be invited even if you don't exist in Synapse. Boo :(

Half-Shot:
Fixes #96

### DIFF
--- a/lib/components/membership-cache.js
+++ b/lib/components/membership-cache.js
@@ -44,7 +44,12 @@ class MembershipCache {
         if (this._membershipMap[roomId] === undefined) {
             this._membershipMap[roomId] = {};
         }
-        this._registeredUsers.add(userId);
+
+        // Bans and invites do not mean the user exists.
+        if (membership === "join" || membership === "leave") {
+            this._registeredUsers.add(userId);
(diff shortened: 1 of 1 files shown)


lib/components/membership-cache.js
```diff
@@ -44,7 +44,12 @@ class MembershipCache {
         if (this._membershipMap[roomId] === undefined) {
             this._membershipMap[roomId] = {};
         }
-        this._registeredUsers.add(userId);
+
+        // Bans and invites do not mean the user exists.
+        if (membership === "join" || membership === "leave") {
```
turt2live: To make the comment true, don't you need this?:
```suggestion
        if (membership === "join" || membership === "leave" || membership === "ban") {
```

Half-Shot: Nope, we are *only* interested in `join` / `leave`. Not `ban` or `invite`.

turt2live: doh, of course. I can read.

PR #104, opened 2019-05-01
Replace .done() with .then/catch

Half-Shot:


turt2live:
Please fix your test failures before asking for review.

PR #105, opened 2019-05-05
Unbreak outcomeFrom

Half-Shot:


turt2live:
Can someone explain what the issue is please?

Half-Shot:
`self.then(func)` doesn't pass `self` as `this` to `func`, which was causing `this.defer` to be undefined.

PR #107, opened 2019-05-17
Create an EventStore for Cadair

### LINKED ISSUE #92: EventStore for storing eventIds against remote ids
and any details about them. Essentially this is just like the other stores. We would also include a source key to tell us if the event came from the remote or matrix side.

Keys required are:
- Matrix Room Id
- Remote Room Id
- Matrix Event Id
- Remote Event Id
- Source
- Data (could be sender or something it we cared about senders)

Half-Shot:
Fixes #92 

Some bridges want to be able to store events ids long term, mapped against their bridged counterparts for things like replies and redactions. This should be a suitable starting point.

### DIFF
--- a/lib/bridge.js
+++ b/lib/bridge.js
@@ -564,7 +564,7 @@ Bridge.prototype.addAppServicePath = function(opts) {
 };
 
 /**
- * Retrieve a connected room store instance.
+ * Retrieve the connected room store instance.
  * @return {?RoomBridgeStore} The connected instance ready for querying.
  */
 Bridge.prototype.getRoomStore = function() {
@@ -572,7 +572,7 @@ Bridge.prototype.getRoomStore = function() {
 };
 
 /**
- * Retrieve a connected user store instance.
+ * Retrieve the connected user store instance.
  * @return {?UserBridgeStore} The connected instance ready for querying.
  */
 Bridge.prototype.getUserStore = function() {
--- a/lib/bridge.js
+++ b/lib/bridge.js
@@ -580,7 +580,7 @@ Bridge.prototype.getUserStore = function() {
 };
 
 /**
- * Retrieve a connected event store instance.
+ * Retrieve the connected event store instance, if one was configured.
  * @return {?EventBridgeStore} The connected instance ready for querying.
  */
 Bridge.prototype.getEventStore = function() {
--- a/lib/bridge.js
+++ b/lib/bridge.js
@@ -66,6 +66,9 @@ const INTENT_CULL_EVICT_AFTER_MS = 1000 * 60 * 15; // 15 minutes
  * @param {(UserBridgeStore|string)=} opts.userStore The user store instance to
  * use, or the path to the user .db file to load. A database will be created if
  * this is not specified.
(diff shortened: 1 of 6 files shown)


turt2live:
don't forget copyright headers

Half-Shot:
Good point :S

lib/bridge.js
```diff
@@ -567,6 +579,14 @@ Bridge.prototype.getUserStore = function() {
     return this._userStore;
 };
 
+/**
+ * Retrieve a connected event store instance.
```
turt2live:
```suggestion
 * Retrieve the connected event store instance, if one was configured.
```

lib/components/event-bridge-store.js
```diff
@@ -0,0 +1,96 @@
+"use strict";
```
turt2live: kill it with fire

also copyright the foundation

lib/models/events/event.js
```diff
@@ -0,0 +1,113 @@
+"use strict";
```
turt2live: death to use strict

also copyright matrix.org foundation thingy

lib/models/events/event.js
```diff
@@ -0,0 +1,113 @@
...
+/**
+ * Create a store event.
+ * @constructor
+ * @param {string} roomId The matrix room ID
+ * @param {string} eventId The matrix event ID
+ * @param {string} remoteRoomId The remote room ID
+ * @param {string} remoteEventId The remote event ID
+ * @param {any} extras Any extra data that may be included with the event.
+ */
+function StoreEvent(roomId, eventId, remoteRoomId, remoteEventId, extras) {
```
turt2live: Can we call this Store**d**Event please? It feels a bit like a typo when reading through the code, despite it being correct as a "store event".

spec/integ/event-bridge-store.spec.js
```diff
@@ -0,0 +1,121 @@
+"use strict";
```
turt2live: strict--

copyright that CIC thing

Cadair:
I can also say I have been able to use this successfully in the slack AS. i.e. matrix-org/matrix-appservice-slack#129 and matrix-org/matrix-appservice-slack#130

Cadair:
@Half-Shot This needs merging before we can release the slack AS

turt2live:
Please avoid force pushes - it makes it harder to review.

lib/components/event-bridge-store.js
```diff
@@ -0,0 +1,107 @@
+/*
```
turt2live: Please maintain consistent whitespace with other projects: https://github.com/matrix-org/matrix-react-sdk/blob/e5059fdf0f82b120c50ad2c9447726b7f9885ce2/src/components/views/settings/Notifications.js#L1-L15

lib/models/events/event.js
```diff
@@ -0,0 +1,124 @@
+/*
```
turt2live: whitespace

spec/integ/event-bridge-store.spec.js
```diff
@@ -0,0 +1,132 @@
+/*
```
turt2live: whitespace

PR #110, opened 2019-05-30
Repair linting

V02460:
Some small changes to make linting work again properly. Fixes vulnerabilites, deprecation warnings, linter errors and a superfluous comma.

- upgrade eslint `^1.2.0`→`^5.16.0`
- rename `.eslintrc`→`.eslintrc.json`
- replace eslint attribute `space-after-keywords` with `keyword-spacing`
- add explicit returns to functions where they are missing to avoid lint erros

Note: I did not fix the linter *warnings* to avoid tainting the blame.

### DIFF
--- a/spec/integ/bridge.spec.js
+++ b/spec/integ/bridge.spec.js
@@ -705,6 +705,7 @@ function mkMockMatrixClient(uid) {
         if (method === "POST" && path === "/register") {
             return client.register(data.user);
         }
+        return undefined;
     });
     client.credentials.userId = uid;
     return client;
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -2,7 +2,7 @@
   "env": {
     "node": true,
     "jasmine": true,
-    "es6": true,
+    "es6": true
   },
   "rules": {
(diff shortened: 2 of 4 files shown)


Half-Shot:
Useful, thanks! I realise that we don't state this anywhere but can you sign off as per https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.rst?

spec/integ/bridge.spec.js
```diff
@@ -705,6 +705,7 @@ function mkMockMatrixClient(uid) {
         if (method === "POST" && path === "/register") {
             return client.register(data.user);
         }
+        return undefined;
```
Half-Shot: I'm surprised this is necessary

V02460: The linter made me do it^^ …But I would have done it the same way anyway as it is more explicit in what is happening. Null(/undefined) values are a pita even if they are in plain sight.

Half-Shot: Fair :)

PR #111, opened 2019-06-13
Enable signaling of bridge errors

V02460:
Adds the ability of signaling permanent errors occuring at the bridge to the SDK as well as automatic sending of these events. If the bridge rejects a room event that was handed to it, the room will automatically be notified of this error.

Mainly two things are added by this PR:
- A new method `signalBridgeError` is added to the `Intent` class.
- Rejection of requests provided to the `onEvent` callback will trigger the new intent.

Bridges should not have to do much work or work at all to profit from the new feature. Every bridge rejecting the request object delivered to it (as should be the case already) will profit. If a bridge swallows errors and simply logs them, code has to be changed to reject the request object properly.

In addition to these changes there is some refactoring and documentation going on, mainly related to the `EventQueue` class.

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,3 +182,22 @@ will not be emitted by `onEvent`. For more information, see the docs.
  * `RemoteRoom` - A representation of a third-party room.
  * `MatrixUser` - A representation of a matrix user.
  * `RemoteUser` - A representation of a third-party user.
+
+
+## Signaling Bridge Errors
+This section applies when you are using `Bridge` and want to notify your users
+about problems while processing their events.
+
+One thing the bridge requires you to do is fulfilling or rejecting the
+`request` promise which is handed to you as argument of the
+`controller.onEvent` callback. When rejecting the promise the `Error` you
+reject with does matter:
+
+- On most `Error`s the bridge will simply notify the homeserver that the event
+  could not be handled right now. The homeserver will then try to resend the
+  message again with exponential backoff.
+- On an `EventNotHandledError` (and all its subtypes) the bridge will declare
+  the event as permanently failed. It will mark it as such by sending a
+  `de.nasnotfound.bridge_error` room event, which will make clients show an
+  error message to their users. The homeserver is signaled that the message was
+  handled and it will not try to resend it.
--- a/README.md
+++ b/README.md
@@ -190,14 +190,12 @@ about problems while processing their events.
 
 One thing the bridge requires you to do is fulfilling or rejecting the
 `request` promise which is handed to you as argument of the
-`controller.onEvent` callback. When rejecting the promise the `Error` you
-reject with does matter:
+`controller.onEvent` callback. When rejecting the promise, the `Error` you
+reject with will indicate to the bridge library how to behave:
 
-- On most `Error`s the bridge will simply notify the homeserver that the event
-  could not be handled right now. The homeserver will then try to resend the
-  message again with exponential backoff.
 - On an `EventNotHandledError` (and all its subtypes) the bridge will declare
   the event as permanently failed. It will mark it as such by sending a
   `de.nasnotfound.bridge_error` room event, which will make clients show an
-  error message to their users. The homeserver is signaled that the message was
-  handled and it will not try to resend it.
+  error message to their users.
+- On all other `Error` types no message is sent to the clients. The bridge
+  still uses the information that the event was handled for queuing purposes.
--- a/lib/bridge.js
+++ b/lib/bridge.js
@@ -768,11 +768,11 @@ Bridge.prototype._onAliasQuery = function(alias) {
 };
 
 // returns a Promise for the request linked to this event for testing.
-Bridge.prototype._onEvent = function(event) {
+Bridge.prototype._onEvent = async function(event) {
     this._updateIntents(event);
     if (this.opts.suppressEcho &&
             this.opts.registration.isUserMatch(event.user_id, true)) {
-        return Promise.resolve();
+        return null;
     }
 
     if (this._roomUpgradeHandler) {
@@ -780,7 +780,7 @@ Bridge.prototype._onEvent = function(event) {
         if (event.type === "m.room.tombstone" && this._roomUpgradeHandler) {
             this._roomUpgradeHandler.onTombstone(event);
             if (this.opts.roomUpgradeOpts.consumeEvent) {
-                return Promise.resolve();
+                return null;
             }
         }
         else if (event.type === "m.room.member" &&
@@ -791,7 +791,7 @@ Bridge.prototype._onEvent = function(event) {
             const isUpgradeInvite = this._roomUpgradeHandler.onInvite(event);
             if (isUpgradeInvite &&
                 this.opts.roomUpgradeOpts.consumeEvent) {
-                return Promise.resolve();
+                return null;
             }
         }
     }
--- a/lib/bridge.js
+++ b/lib/bridge.js
@@ -873,7 +873,7 @@ Bridge.prototype._handleEventError = function(event, error) {
         event.event_id,
         "Discord",
         error.reason,
-        this._getUserRegex()
+        this._getUserRegex(),
     );
 };
 
@@ -884,8 +884,7 @@ Bridge.prototype._handleEventError = function(event, error) {
  */
 Bridge.prototype._getUserRegex = function() {
     const reg = this.opts.registration;
-    const userRegexes = reg.namespaces["users"].map(o => o.regex);
-    return userRegexes.join("|");
+    return reg.namespaces["users"].map(o => o.regex);
 };
 
 Bridge.prototype._updateIntents = function(event) {
--- a/lib/bridge.js
+++ b/lib/bridge.js
@@ -813,9 +813,15 @@ Bridge.prototype._onEvent = async function(event) {
 };
 
 /**
- * Returns a promise that is like `promise` but adheres to all necessary limits.
+ * Restricts the promise according to the bridges `perRequest` setting.
+ *
+ * `perRequest` enabled:
+ *     Returns a promise similar to `promise`, with the addition of it only
+ *     resolving after `request`.
+ * `perRequest` disabled:
+ *     Returns the promise unchanged.
  */
-Bridge.prototype._limited = function(promise, request) {
+Bridge.prototype._limited = async function(promise, request) {
     // queue.perRequest controls whether multiple request can be processed by
     // the bridge at once.
     if (this.opts.queue.perRequest) {
--- a/lib/bridge.js
+++ b/lib/bridge.js
@@ -30,6 +30,16 @@ const INTENT_CULL_CHECK_PERIOD_MS = 1000 * 60; // once per minute
 // How long a given Intent object can hang around unused for.
 const INTENT_CULL_EVICT_AFTER_MS = 1000 * 60 * 15; // 15 minutes
 
+/* Returns a new promise which behaves like `main`, but holds settling back
+ * until `prolonger` was settled.
+ * @param {Promise} main The promise to mimick.
+ * @param {Promise} prolonger The promise to wait for.
+ * @return {Promise} A new promise behaving like main, but waiting for prolonger.
+ */
+function prolong(main, prolonger) {
+    return Promise.all([main, prolonger.reflect()]).then(([m, _]) => m);
+}
+
 /**
  * @constructor
  * @param {Object} opts Options to pass to the bridge
@@ -793,51 +803,46 @@ Bridge.prototype._onEvent = function(event) {
         }
     }
 
-    var request = this._requestFactory.newRequest({ data: event });
-    var context = new BridgeContext({
-        sender: event.user_id,
-        target: event.state_key,
-        room: event.room_id
-    });
-    var data = {
+    const request = this._requestFactory.newRequest({ data: event });
+    const contextReady = this._getBridgeContext(event);
+    const dataReady = contextReady.then(c => ({
         request: request,
-        context: context
-    };
-
-    var promise;
-    if (this.opts.disableContext) {
-        promise = Promise.resolve();
-        data.context = null;
-    }
-    else {
-        promise = context.get(this._roomStore, this._userStore);
-    }
+        context: c
+    }));
 
     if (this.opts.queue.perRequest) {
-        promise = Promise.settle([
-            promise,
-            this._prevRequestPromise
-        ]);
+        dataReady = prolong(dataReady, this._prevRequestPromise)
         this._prevRequestPromise = request.getPromise();
     }
 
-    this._queue.push(event, data, promise);
+    this._queue.push(event, dataReady);
     this._queue.consume();
     return request.getPromise();
 };
 
 Bridge.prototype._onConsume = function(err, data) {
-    if (!err) {
-        this.opts.controller.onEvent(data.request, data.context);
+    if (err) {
+        this.opts.controller.onLog("onEvent failure: " + err);
         return;
     }
-    if (!this.opts.controller.onLog) {
-        return;
-    }
-    this.opts.controller.onLog(
-        "onEvent failure: " + err
-    );
+
+    this.opts.controller.onEvent(data.request, data.context);
 };
+
+Bridge.prototype._getBridgeContext = function(event) {
+    if (this.opts.disableContext) {
+        return Promise.resolve(null);
+    }
+
+    const context = new BridgeContext({
+        sender: event.user_id,
+        target: event.state_key,
+        room: event.room_id
+    });
+    const contextReady = context.get(this._roomStore, this._userStore)
+
+    return contextReady.return(context);
+}
 
 Bridge.prototype._updateIntents = function(event) {
     if (event.type === "m.room.member") {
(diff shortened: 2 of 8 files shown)


Half-Shot:
Thanks for submitting the PR. Could you give a rough changelog in the PR description for the viewers at home?

lib/bridge.js
```diff
@@ -972,57 +963,101 @@ function queueAlgorithm(event) {
...
-        self.consumeFn(null, entry.data);
-        self._takeNext(identifier);
-    }, function(e) {
-        self.consumeFn(e, null);
-        self._takeNext(identifier);
-    });
-};
+}
+
+class EventQueuePerRoom extends EventQueue {
```
Half-Shot: Thank you for using ES6 classes, it's quite refreshing :). Could the EventQueue stuff be taken to a seperate EventQueue.js file, as it makes for easier reviewing

lib/bridge.js
```diff
@@ -964,29 +969,33 @@ function queueAlgorithm(event) {
 }
 
 class EventQueue {
```
Half-Shot: For each of these classes it would be good to have a jsdoc explaining what they do.

lib/bridge.js
```diff
@@ -964,29 +969,33 @@ function queueAlgorithm(event) {
...
+            //  events: [ {dataReady: } ],
             //  consuming: true|false
             // }
         };
         this.consumeFn = consumeFn;
     }
 
-    push(event, data, promise) {
-        var identifier = this.type === "per_room" ? event.room_id : "none";
+    push(event, dataReady) {
```
Half-Shot: Same for each of these functions, a [JSDoc](https://en.wikipedia.org/wiki/JSDoc) line would go a long way :)

lib/bridge.js
```diff
@@ -773,52 +783,47 @@ Bridge.prototype._onEvent = function(event) {
...
-    }
+        context: c
+    }));
 
     if (this.opts.queue.perRequest) {
-        promise = Promise.settle([
-            promise,
-            this._prevRequestPromise
-        ]);
+        dataReady = prolong(dataReady, this._prevRequestPromise)
```
Half-Shot: Out of interest, is there anytthing wrong with `dataReady = this._prevRequestPromise.finally(() => dataReady)`?

`finally` will fire no matter the outcome, and if the promise has already resolved/rejected, it will still fire.

lib/bridge.js
```diff
@@ -773,52 +783,47 @@ Bridge.prototype._onEvent = function(event) {
...
-    this._queue.push(event, data, promise);
+    this._queue.push(event, dataReady);
     this._queue.consume();
     return request.getPromise();
 };
 
 Bridge.prototype._onConsume = function(err, data) {
-    if (!err) {
-        this.opts.controller.onEvent(data.request, data.context);
+    if (err) {
```
Half-Shot: Do we not need to check for `this.opts.controller.onLog` ?

lib/bridge.js
```diff
@@ -1000,20 +1009,15 @@ class EventQueue {
     };
 
     _takeNext(identifier) {
-        var self = this;
         var events = this._queues[identifier].events;
```
Half-Shot: Cheeky request: Make these `const`/`let`?

lib/bridge.js
```diff
@@ -29,6 +29,16 @@ const INTENT_CULL_CHECK_PERIOD_MS = 1000 * 60; // once per minute
...
 const INTENT_CULL_EVICT_AFTER_MS = 1000 * 60 * 15; // 15 minutes
 
+/* Returns a new promise which behaves like `main`, but holds settling back
+ * until `prolonger` was settled.
+ * @param {Promise} main The promise to mimick.
+ * @param {Promise} prolonger The promise to wait for.
+ * @return {Promise} A new promise behaving like main, but waiting for prolonger.
+ */
+function prolong(main, prolonger) {
+    return Promise.all([main, prolonger.reflect()]).then(([m, _]) => m);
```
Half-Shot: Could this not be achieved by `prolonger.finally(() => main)`?

lib/errors.js
```diff
@@ -0,0 +1,93 @@
+"use strict";
```
Half-Shot: Use strict is old hat at this point, you are probably safe to remove it.

lib/bridge.js
V02460: I am new to the JS Promise world. To me it looks like some flattening magic is going on in your example code, but I think it does the right thing :)

lib/bridge.js
V02460: Afaics it is accounted for in the constructor:

https://github.com/matrix-org/matrix-appservice-bridge/blob/f5fbf9d9c3f36dbad731f6151e234f4ccd6ce705/lib/bridge.js#L156-L163

V02460:
Changed PR from “draft” to “ready for review” as [Travis CI does currently not work with draft PRs](https://travis-ci.community/t/draft-pull-requests-not-being-built/2434).

lib/bridge.js
Half-Shot: Oh aren't I clever. Yep, that's cool.

lib/components/intent.js
```diff
@@ -535,6 +535,49 @@ Intent.prototype.setPresence = function(presence, status_msg=undefined) {
...
+ * @return {Promise}
+ */
+Intent.prototype.signalBridgeError = function(
+    roomId,
+    eventId,
+    network_name,
+    reason,
+    affected_users
+) {
+    if (!affected_users) {
```
Half-Shot: Interesting. Could we save some bytes here and have affected users be `["*"]`. It feels a bit nicer than trying to fetch all the current users.

lib/components/intent.js
```diff
@@ -535,6 +535,49 @@ Intent.prototype.setPresence = function(presence, status_msg=undefined) {
...
+    }
+
+    return this.sendEvent(
+        roomId,
+        "m.room.bridge_error",
+        {
+            "network_name": network_name,
+            "reason": reason,
+            "affected_users": affected_users,
+            "m.relates_to": {
```
Half-Shot: +1 for relates to! :)

lib/components/intent.js
```diff
@@ -535,6 +535,49 @@ Intent.prototype.setPresence = function(presence, status_msg=undefined) {
...
+    affected_users
+) {
+    if (!affected_users) {
+        const room = this.client.getRoom(roomId)
+        affected_users = room.getJoinedMembers().map(m => m.userId)
+    }
+
+    return this.sendEvent(
+        roomId,
+        "m.room.bridge_error",
```
Half-Shot: When a event type is unspecced and in a testing state, we prefer to use a custom namespace like `com.example.room.bridge_error` so it doesn't clash with any future event types. You should do this here.

V02460: Do you mean `com.example` literally or should I use my own domain name?

Half-Shot: Use your own.

spec/unit/intent.spec.js
```diff
@@ -19,7 +19,7 @@ describe("Intent", function() {
         var clientFields = [
             "credentials", "joinRoom", "invite", "leave", "ban", "unban",
             "kick", "getStateEvent", "setPowerLevel", "sendTyping", "sendEvent",
-            "sendStateEvent", "setDisplayName", "setAvatarUrl"
+            "sendStateEvent", "setDisplayName", "setAvatarUrl", "getRoom"
```
Half-Shot: where is getRoom used?

lib/errors.js
```diff
@@ -0,0 +1,100 @@
...
+    constructor(...args) {
+        default_message(args, "The event is not known to the bridge");
+        super(...args);
+        this.name = "EventUnknownError";
+        this.reason = "m.event_unknown";
+    }
+}
+
+module.exports = {
+    wrap,
```
Half-Shot: `wrap` might be too generic a name :/

lib/errors.js
```diff
@@ -0,0 +1,100 @@
...
+        args[0] = msg;
+   }
+}
+
+/**
+ * Base Error for when the bride can not handle the event.
+ */
+class EventNotHandledError extends Error {
+    constructor(...args) {
+        default_message(args, "The event could not be handled by the bridge");
```
Half-Shot: You can also do `args[0] = args[0] || "This is my default message"`

The `||` operator is magical :)

lib/errors.js
```diff
@@ -0,0 +1,100 @@
...
+        append_msg = old_error.toString();
+    }
+    new_error.message += ":\n" + append_msg;
+    return new_error;
+}
+
+/**
+ * Sets the default message for the given error arguments.
+ */
+function default_message(args, msg) {
```
Half-Shot: As per the other comment, we can drop this.

lib/errors.js
```diff
@@ -0,0 +1,100 @@
+/**
+ * Append the old error message to the new one and keep its stack trace.
+ * Example:
+ *     throw wrap(e, HighLevelError, "This error is more specific");
+ */
+function wrap(
+    old_error,
```
Half-Shot: We should lint this, but var names should be camelCase

lib/components/intent.js
```diff
@@ -535,6 +535,47 @@ Intent.prototype.setPresence = function(presence, status_msg=undefined) {
...
+    eventId,
+    network_name,
+    reason,
+    affected_users
+) {
+    return this.sendEvent(
+        roomId,
+        "de.nasnotfound.bridge_error",
+        {
+            "network_name": network_name,
```
Half-Shot: You don't need to quote JSON in Javascript, it's better without. Also you can use shorthand here for nearly all of these names like:

```
{
   network_name,
   reason,
   affected_users, //NOTE: You may want to validate this one
   "m.relates_to": { stuff in here ...}
}
```

lib/bridge.js
```diff
@@ -1074,7 +1056,7 @@ BridgeContext.prototype.get = function(roomStore, userStore) {
         if (mxSender) {
             self.senders.matrix = mxSender;
         }
-    });
+    }).catch(e => {throw wrap(e, Error, "Could not retrieve bridge context");});
```
Half-Shot: My advice for this feature is to disable it by default and stick a config flag in the constructor at the top like we do for things like room upgrades. It means bridges that rely on previous behaviour will not explode, but it will still allow them to opt in with a one liner :)

lib/bridge.js
```diff
@@ -773,61 +776,92 @@ Bridge.prototype._onEvent = function(event) {
...
+        this._getUserRegex()
     );
 };
 
+/**
+ * Returns a regex matching all users of the bridge.
+ *
+ * @return {string} Super regex composed of all user regexes.
+ */
+Bridge.prototype._getUserRegex = function() {
```
Half-Shot: I'd expect a RegExp object here :)

lib/bridge.js
```diff
@@ -773,61 +776,92 @@ Bridge.prototype._onEvent = function(event) {
...
+    const contextReady = context.get(this._roomStore, this._userStore)
+
+    return contextReady.return(context);
+}
+
+Bridge.prototype._handleEventError = function(event, error) {
+    if (!error instanceof EventNotHandledError) {
+        error = wrap(error, InternalError);
+    }
+    this._botIntent.signalBridgeError(
```
Half-Shot: Do we need to handle cases where the bot may not be in the room?

lib/bridge.js
```diff
@@ -773,61 +776,92 @@ Bridge.prototype._onEvent = function(event) {
...
+        this.opts.controller.onLog("onEvent failure: " + err);
         return;
     }
-    if (!this.opts.controller.onLog) {
-        return;
+
+    this.opts.controller.onEvent(data.request, data.context);
+};
+
+Bridge.prototype._getBridgeContext = function(event) {
```
Half-Shot: async function?

lib/bridge.js
```diff
@@ -773,61 +776,92 @@ Bridge.prototype._onEvent = function(event) {
...
-        room: event.room_id
-    });
-    var data = {
+    const request = this._requestFactory.newRequest({ data: event });
+    const contextReady = this._getBridgeContext(event);
+    const dataReady = contextReady.then(c => ({
         request: request,
-        context: context
-    };
+        context: c
```
Half-Shot: Eek, I'd rather you used shorthand here so ```{request, context}```

spec/unit/intent.spec.js
V02460: Nowhere – it's a leftover from the old implementation.

lib/errors.js
V02460: Oh…we're not doing Python here :smile: Python would make you suffer for that^^°

I'd argue to keep the method, as it makes the intent clearer. (I would still clean up `default_message` then.)

lib/bridge.js
V02460: The more you know :rainbow:

spec/unit/intent.spec.js
Half-Shot: Can probably drop this then :)

lib/bridge.js
V02460: I am not sure which feature you are talking about here. With this line an Error is thrown, without this line an Error is thrown :thinking: Could you please elaborate a bit more?

lib/bridge.js
V02460: Well, it does return a Promise… If I added the `asyc` keyword, wouldn't that make it return `Promise<Promise<BridgeContext|null>>`?

lib/bridge.js
V02460: Yes. As this is more a design question and currently discussed in context of the MSC I might want to delay it here.

lib/bridge.js
V02460: You suggest changing the name to something like `_getUserRegexString`?

lib/errors.js
V02460: https://github.com/matrix-org/matrix-appservice-bridge/pull/115

lib/bridge.js
Half-Shot: Yep, sorry. It was tricky to find the best line to use for context.

Automatically sending signal events into the room upon failure is probably not something we want on by default (which is what I assume this line is doing?). The reason for this is that it might do unexpected and unwanted things for users who are just upgrading their bridge version. Could we put automatic signaling behind a constructor flag?

lib/bridge.js
Half-Shot: I mean that this function should return a RegExp object, not a string.

lib/bridge.js
Half-Shot: You can put in a TODO line here just to explain that it's in progress.

lib/bridge.js
Half-Shot: JS is smart, and it doesn't wrap promises like that but chains them (so you will never have to handle a promise inside a promise). Making the func `async` means you can `return null;` rather than `return Promise.resolve(null);`

lib/bridge.js
V02460: Just to turn it back into a string? O_o The method is only there to get the string representation of the regex for the event.

lib/bridge.js
V02460: Oh, I get it, you are talking about this whole feature :D

https://github.com/matrix-org/matrix-appservice-bridge/blob/dcd323e17a6d0314f09756c21f0a90232b538601/lib/bridge.js#L794

This is your feature switch, essentially. Only if the bridge rejects with an `EventNotHandledError` or one of its subtypes, the behavior is triggered. By that, bridges which do not change anything, don't get anything new.

The specific line referenced by you is pretty close to a no-op: In the case of an Error already thrown, it catches it and rethrows a new one, just with another error message.

lib/bridge.js
Half-Shot: My bad, you're not using it as a regex anywhere here but just to stick inside the event. Got it.

lib/bridge.js
Half-Shot: That's very clever and I totally misunderstood it. Makes more sense to me now :)

Half-Shot:
So this all looks good after https://github.com/matrix-org/matrix-appservice-bridge/pull/111#discussion_r304776393 gets resolved. Can we get a quick how-to use section in README.md like we do for other components so other bridge developers can understand? This should supplement the jsdoc

README.md
```diff
@@ -179,3 +179,22 @@ will not be emitted by `onEvent`. For more information, see the docs.
...
  * `RemoteUser` - A representation of a third-party user.
+
+
+## Signaling Bridge Errors
+This section applies when you are using `Bridge` and want to notify your users
+about problems while processing their events.
+
+One thing the bridge requires you to do is fulfilling or rejecting the
+`request` promise which is handed to you as argument of the
+`controller.onEvent` callback. When rejecting the promise the `Error` you
```
Half-Shot: comma

```suggestion
`controller.onEvent` callback. When rejecting the promise, the `Error` you
```

README.md
```diff
@@ -179,3 +179,22 @@ will not be emitted by `onEvent`. For more information, see the docs.
...
+reject with does matter:
```
Half-Shot: Matter felt a little...ambiguous?

```suggestion
reject with will indicate to the bridge library how to behave:
```

README.md
```diff
@@ -179,3 +179,22 @@ will not be emitted by `onEvent`. For more information, see the docs.
...
+
+- On most `Error`s the bridge will simply notify the homeserver that the event
```
Half-Shot: I would reorder this list, and rephrase this to say "On all other Error types, the bridge will..."

README.md
```diff
@@ -179,3 +179,22 @@ will not be emitted by `onEvent`. For more information, see the docs.
...
+  could not be handled right now. The homeserver will then try to resend the
+  message again with exponential backoff.
```
Half-Shot: This sounds strange to me. Would an Error cause the bridge to emit a non-200 status code to the homeserver in order to make the homeserver resend it. It sounds dangerous to me to start feeding bridges the same events, as I believe there may be bridges at present which return `Error` for minor things, while the event may have been processed.

I must have missed this in my previous review, but I would urge that Errors are handled the same way as successes and just use your custom Error types to influence bridge logic.

V02460: Reread some docs and code. I got that part wrong >.< The AS (matrix-appservice-node) will always respond with a 200 as soon as the callback finishes (regardless of what its returned promise is). The event is then irrevocable the problem of the bridge.

That makes me want to reconsider some things about the MSC. Even with that misunderstanding, I think the code is still ok in this form for now.

Half-Shot: >Reread some docs and code. I got that part wrong >.< The AS (matrix-appservice-node) will always respond with a 200 as soon as the callback finishes (regardless of what its returned promise is). The event is then irrevocable the problem of the bridge.

Ah good. I was worried you may have changed this and I missed it. Anyway that's cool. I guess you can update this README statement and then the code should be good :)

lib/errors.js
V02460: https://github.com/Half-Shot/matrix-appservice-discord/pull/512#discussion_r304373100

Half-Shot:
Looks like there are a few linting warnings we can clean up, and istanbul is not pleased about our async syntax :(

V02460:
I cleaned up linter warnings in all files I touched and did https://github.com/matrix-org/matrix-appservice-bridge/pull/116 for the issue with Istanbul.

lib/bridge.js
```diff
@@ -788,66 +791,97 @@ Bridge.prototype._onEvent = function(event) {
...
+        request
+        .getPromise()
+        .catch(EventNotHandledError, e => this._handleEventError(event, e))
+    );
+};
+
+/**
+ * Returns a promise that is like `promise` but adheres to all necessary limits.
+ */
+Bridge.prototype._limited = function(promise, request) {
```
Half-Shot: Is this always returning a promise?

lib/bridge.js
```diff
@@ -788,66 +791,97 @@ Bridge.prototype._onEvent = function(event) {
...
 
+    return (
+        request
+        .getPromise()
+        .catch(EventNotHandledError, e => this._handleEventError(event, e))
+    );
+};
+
+/**
+ * Returns a promise that is like `promise` but adheres to all necessary limits.
```
Half-Shot: This needs to be less vague. What are "all necessary limits"?

lib/components/event-queue.js
```diff
@@ -0,0 +1,160 @@
...
+ * Dummy EventQueue for which no queue is utilized.
+ *
+ * Every event is handled as soon as its data is available.
+ */
+class EventQueueNone extends EventQueue {
+    constructor(consumeFn) {
+        super("none", consumeFn);
+    }
+
+    push(event, data, promise) {
```
Half-Shot: as far as I can tell, you never pass a third value into this function anywhere else in the PR which makes it go bang.

lib/bridge.js
V02460: Made it async :)

lib/bridge.js
V02460: Hope its clearer now.

Half-Shot:
Okay, have tested this on matrix-appservice-irc and it does the right thing so I'll merge it in \o/

PR #112, opened 2019-06-27
Escape user IDs correctly

### LINKED ISSUE #109: escapeUserId misses [ and ]

Half-Shot:
This fixes #109. In addition, this PR adds a global variable to disable userid escaping for bridges that don't support it yet.

### DIFF
--- a/lib/models/users/matrix.js
+++ b/lib/models/users/matrix.js
@@ -59,7 +59,7 @@ MatrixUser.prototype.set = function(key, val) {
     this._data[key] = val;
 };
 
-/**u
+/**
  * Get the data value for the given key.
  * @param {string} key An arbitrary bridge-specific key.
  * @return {*} Stored data for this key. May be undefined.
--- a/lib/models/users/matrix.js
+++ b/lib/models/users/matrix.js
@@ -80,8 +80,9 @@ MatrixUser.prototype.serialize = function() {
  * Grammar taken from: https://matrix.org/docs/spec/appendices.html#identifier-grammar
  */
 MatrixUser.prototype.escapeUserId = function() {
-    // Currently Matrix accepts / in the userId, although going forward it will be removed.
-    const badChars = new Set(this.localpart.replace(/([A-z0-9]|-|\.|=|_)+/g, ""));
+    // NOTE: Currently Matrix accepts / in the userId, although going forward it will be removed.
+    // NOTE: We also allow uppercase for the time being.
+    const badChars = new Set(this.localpart.replace(/([A-Z]|[a-z]|[0-9]|-|\.|=|_)+/g, ""));
     let res = this.localpart;
     badChars.forEach((c) => {
         const hex = c.charCodeAt(0).toString(16).toLowerCase();
--- a/lib/models/users/matrix.js
+++ b/lib/models/users/matrix.js
@@ -5,9 +5,10 @@
  * @constructor
  * @param {string} userId The user_id of the user.
  * @param {Object=} data Serialized data values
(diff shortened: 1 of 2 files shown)


lib/models/users/matrix.js
```diff
@@ -58,7 +59,7 @@ MatrixUser.prototype.set = function(key, val) {
     this._data[key] = val;
 };
 
-/**
+/**u
```
turt2live: u

lib/models/users/matrix.js
```diff
@@ -75,13 +76,15 @@ MatrixUser.prototype.serialize = function() {
...
 /**
  * Make a userId conform to the matrix spec using QP escaping.
  * Grammar taken from: https://matrix.org/docs/spec/appendices.html#identifier-grammar
  */
 MatrixUser.prototype.escapeUserId = function() {
-    // Currently Matrix accepts / in the userId, although going forward it will be removed.
-    const badChars = new Set(this.localpart.replace(/([A-z0-9]|-|\.|=|_)+/g, ""));
+    // NOTE: Currently Matrix accepts / in the userId, although going forward it will be removed.
+    // NOTE: We also allow uppercase for the time being.
+    const badChars = new Set(this.localpart.replace(/([A-Z]|[a-z]|[0-9]|-|\.|=|_)+/g, ""));
```
turt2live: we should remove characters which aren't allowed, such as uppercase. I'm pretty sure Synapse has started failing to register uppercase usernames by now.

lib/models/users/matrix.js
```diff
@@ -83,13 +83,17 @@ MatrixUser.prototype.serialize = function() {
...
-    // NOTE: We also allow uppercase for the time being.
-    const badChars = new Set(this.localpart.replace(/([A-Z]|[a-z]|[0-9]|-|\.|=|_)+/g, ""));
+    const badChars = new Set(this.localpart.replace(/([a-z]|[0-9]|-|\.|=|_)+/g, ""));
     let res = this.localpart;
     badChars.forEach((c) => {
-        const hex = c.charCodeAt(0).toString(16).toLowerCase();
+        const code = c.charCodeAt(0);
+        const hex = code.toString(16).toLowerCase();
+        if (code < 65 || code > 90) {
+            // Alpha codes do not need escaping.
```
turt2live: ... but should be lowercased.

spec/unit/matrix-user.spec.js
```diff
@@ -6,8 +6,8 @@ describe("MatrixUser", function() {
             [
                 new MatrixUser("@test:localhost", null, false),
                 new MatrixUser("@42:localhost", null, false),
-                new MatrixUser("@Test42:localhost", null, false),
-                new MatrixUser("@A=Good-set.of_chars:localhost", null, false)
```
turt2live: which means keeping this test to result in a lowercase version

lib/models/users/matrix.js
Half-Shot: are we lowercasing them rather than escaping them?

turt2live: I don't see why not. User IDs are semi-opaque anyways, so Bob and bob are just going to have to get along.

Half-Shot: I can see that causing lots of unexpected conflicts :c

turt2live: tbh I see anything we do here as creating unexpected conflicts. See also: IRC bridge creating duplicate users, resulting in the bridge doing it's own escaping.

Maybe we just keep uppercase as allowed (sorry) and never visit this code again.

PR #116, opened 2019-07-19
Require ECMAScript 2017

V02460:


Half-Shot:
For reference, this would essentially bump our minimum required node version to node 8. Node 8 is nearly out of support so this is more than fine :).

PR #122, opened 2019-07-25
Add warnings to explain that signalling bridge errors is an unstable feature

Half-Shot:


Half-Shot:
It's a genuine complaint, and my bad for not remembering to do this stuff. Thanks for shouting

turt2live:
for archaeology: regressed by https://github.com/matrix-org/matrix-appservice-bridge/pull/111

PR #123, opened 2019-07-25
Bridge Error: Check if AS user regex has domain part

V02460:
This will prevent the bridge from sending a bridge error by rejecting `unstableSignalBridgeError` if the AS user regex does not contain a local part.


V02460:
Currently this would always log an error when a bridge error event should have been sent. Doing it this way is the cautious way as it doesn't affect bridges not using the feature. When requiring a domain part really should be a global requirement then matrix-appservice-node should implement such a check as well.

Half-Shot:
I think we're unlikely to work on this feature anytime soon. Sorry :(

PR #126, opened 2019-08-07
Export and cleanup BridgeContext

V02460:
The `BridgeContext` class is part of the public interface of the bridge (`controller.onEvent`), so it should be properly exported.

The PR also refactors the class to something appearing to be from this century.

### DIFF
--- a/lib/bridge.js
+++ b/lib/bridge.js
@@ -1062,17 +1062,17 @@ class BridgeContext {
         this.senders = {
             matrix: new MatrixUser(ctx.sender),
             remote: null,
-            remotes: []
+            remotes: [],
         };
         this.targets = {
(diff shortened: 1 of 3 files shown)


lib/bridge.js
```diff
@@ -1103,6 +1047,78 @@ BridgeContext.prototype.get = function(roomStore, userStore) {
  * @property {?RemoteRoom} rooms.remote The first linked remote room: remotes[0]
  * @property {RemoteRoom[]} rooms.remotes The linked remote rooms for this event
  */
+class BridgeContext {
```
Half-Shot: If we're going to export it, put it inside it's own file :innocent:

V02460:
Forgot the new file :sweat_smile:

PR #127, opened 2019-08-07
Catch all request errors before returning from _onEvent

V02460:
Catches the error of the rejected request promise and logs it as responsible for dropping the event. By this `_onEvent` will never return a rejected promise.


Half-Shot:
Seems sensible :)

PR #141, opened 2019-11-14
Add note: replace localhost with server_name

marceltransier:
IMHO it was not clear that localhost is the servername.
(Because in the note which says that the AS user must be invited if the room ist non-public the example userid is `@slackbot:domain` not `:localhost`)


Half-Shot:
Sure, sounds reasonable. Can you sign off as per https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.rst ?

marceltransier:
Signed-off-by: Marcel Transier <me@marceltransier.de>

PR #145, opened 2019-12-04
Remove usages of the v1 event format keys

Half-Shot:
Which in practice is just `user_id`.

Will also depend on https://github.com/matrix-org/matrix-appservice-node/pull/20, when it lands.

~This will require a version bump to 2.0 as existing bridges may depend on the user_id behavior.~ Actually, it won't. Synapse will send bridges V2 events regardless of the version of this library, so meh.


Half-Shot:
ikr

PR #147, opened 2019-12-26
Modernize intent.js and its spec

V02460:
Some housecleaning in preparation for the *bridge information* PR.


Half-Shot:
This is looking good, thanks. I had planned to move this to TS, but ES6 classes help that effort anyway :).

Half-Shot:
Superseded by #185 (thanks for having a go)

PR #150, opened 2020-03-08
Drain the queue after setting up logger

halkeye:
If you log before loggers are setup, but never actually log to that logger again, it just never logs.

Quick patch to drain the log queue as soon as things are setup.

It also could be done on setLogger


halkeye:
@Half-Shot it looks like you don't watch the repo, so to be safe gonna ping you as well

halkeye:
Rebased and signed

PR #151, opened 2020-03-08
use winston.format.splat instead of util.inspect

halkeye:
Not a lot of differences, but winston's splat is a bit more powerful than straight up utils.inspect, and maintained by others

see https://github.com/winstonjs/logform/blob/master/splat.js


halkeye:
meh conflicts, and i didn't sign it. closing it.

PR #155, opened 2020-04-07
Fix bug on PL events without `events` property

martindale:
Ran into this while working on a bridge of my own, thanks to `@david:vovo.id.au` in the Matrix Bridging room for confirming.


Half-Shot:
Thanks for the contrib, would you mind signing off as per https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md ?

jaller94:
This became obsolete by #185.

The new lines are: https://github.com/matrix-org/matrix-appservice-bridge/blob/7df13723346028763958a398602d50877e4eb15c/src/components/intent.ts#L780-L786
My solution even accounts for default_state being undefined or any of these not being numbers.

PR #156, opened 2020-04-10
fix: wrap is used, but not imported

halkeye:


Half-Shot:
Thanks, would you mind signing off your changes as per https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md

halkeye:
Updated and rebased

I sorta thought the github gui would do this :)

PR #157, opened 2020-04-14
Queue /state requests from StateLookup

Half-Shot:
This PR converts the `StateLookup` class to Typescript. It also adds in a queue function to ensure we do not overload the homeserver.


src/components/state-lookup.ts
```diff
@@ -0,0 +1,188 @@
...
+        const r = this.dict[roomId] = this.dict[roomId] || {};
+        if (r.syncPromise) {
+            return r.syncPromise;
+        }
+        r.events = {};
+        r.syncPromise = (async () => {
+            while (true) {
+                try {
+                    const events: StateLookupEvent[] = await this.lookupQueue.add(
+                        () => this._client.roomState(roomId)
```
jaller94, line 129: A returning arrow function inside
a function that is awaited inside
a try block inside
a while(true) block inside
an async function that is directly called inside
a Bluebird.resolve().
What does that even give me?

The complexity:
giphub

src/components/state-lookup.ts
```diff
@@ -0,0 +1,188 @@
...
+        (opts.eventTypes || []).forEach((t) => {
+            this.eventTypes[t] = true;
+        });
+    }
+
+    /**
+     * Get a stored state event.
+     * @param {string} roomId
+     * @param {string} eventType
+     * @param {string=} stateKey If specified, this function will return either
```
jaller94, line 85: What does `string=` mean and why are these type annotation kept in addition with TypeScript?

src/components/state-lookup.ts
```diff
@@ -0,0 +1,188 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+import Bluebird from "bluebird";
+import PQueue from "p-queue";
+
+interface StateLookupOpts {
+    client: any;
```
jaller94, line 19:
```suggestion
    client: any; //TODO: Needs to be MatrixClient (once that becomes TypeScript)
```

src/components/state-lookup.ts
Half-Shot: For now it's just so the docs continue to be generated successfully. In a later PR we can look to clean these up.

Half-Shot:
@jaller94 I forgot to commit the changes that would hopefully simplify this a bit. Sadly, one of the tests still fails.

spec/unit/state-lookup.spec.js
```diff
@@ -82,7 +82,9 @@ describe("StateLookup", function() {
                     // the catch => nextTick magic.
                     var p = Promise.reject(new Error("network error"));
                     p.catch(function(err) {
+                        console.log("WOOF");
```
jaller94, line 85: 🍖

spec/unit/state-lookup.spec.js
```diff
@@ -82,7 +82,6 @@ describe("StateLookup", function() {
                     // the catch => nextTick magic.
                     var p = Promise.reject(new Error("network error"));
                     p.catch(function(err) {
-                        console.log("WOOF");
                         process.nextTick(function() {
                             console.log("TICK!");
```
jaller94, line 86: Another console.log for debugging.

spec/unit/state-lookup.spec.js
```diff
@@ -157,18 +158,15 @@ describe("StateLookup", function() {
...
-            p.then(function() {
-                expect(
-                    lookup.getState("!foo:bar", "m.room.name", "").content.name
-                ).toEqual("Bar");
-                done();
-            });
+            await p;
+            expect(
+                lookup.getState("!foo:bar", "m.room.name", "").content.name
+            ).toEqual("Bar");
```
jaller94, line 169: Let's take a minute and appreciate how much cleaner this code looks now. 👏
Also, on an error, the test will fail instead of time out. 👏👏

Half-Shot: :)

PR #158, opened 2020-04-16
Authenticate requests to bridge endpoints

Half-Shot:
This fixes a bug where we were previously not checking the hsToken on several bridge endpoints.

### DIFF
--- a/src/bridge.js
+++ b/src/bridge.js
@@ -387,9 +387,6 @@ Bridge.prototype._customiseAppservice = function() {
             method: "POST",
             path: "/_bridge/roomLinkValidator/reload",
             handler: (req, res) => {
-                if (!this._requestCheckToken(req, res)) {
-                    return;
-                }
                 try {
                     // Will use filename if provided, or the config
                     // one otherwised.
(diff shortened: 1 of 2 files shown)


src/bridge.js
```diff
@@ -141,6 +141,9 @@ const INTENT_CULL_EVICT_AFTER_MS = 1000 * 60 * 15; // 15 minutes
  * @param {boolean=} opts.roomLinkValidation.triggerEndpoint Enable the endpoint
  * to trigger a reload of the rules file.
  * Default: false
+ * @param {string} opts.authenticateThirdpartyEndpoints Should the bridge authenticate
+ * requests to thirdparty endpoints. This is false by default to be backwards compatible
```
jaller94, line 145: nitpick
```suggestion
 * requests to third party endpoints. This is false by default to be backwards-compatible
```

src/bridge.js
```diff
@@ -444,16 +445,10 @@ Bridge.prototype._customiseAppserviceThirdPartyLookup = function(lookupControlle
 
         this.addAppServicePath({
             method: "GET",
-            path: "/_matrix/app/:ver/thirdparty/protocol/:protocol",
+            path: "/_matrix/app/:version/thirdparty/protocol/:protocol",
```
jaller94, line 448: Why does this one not need the `access_token` to be deleted like most other AppServicePaths?
Should the `delete req.query.access_token;` be moved to a more central function?

Half-Shot: Because it's not using req.query in the function below. We only want to remove it for the two functions which use the whole of req.query.

The reason we use the whole of `req.query` is because of this unfortunate API shape: https://matrix.org/docs/spec/application_service/r0.1.2#get-matrix-app-v1-thirdparty-user-protocol

PR #160, opened 2020-04-16
Limit versions to v1/unstable

Half-Shot:
Missed from #158. Ensure we only allow v1 or unstable requests to these endpoints.

### DIFF
--- a/src/bridge.js
+++ b/src/bridge.js
@@ -445,7 +445,7 @@ Bridge.prototype._customiseAppserviceThirdPartyLookup = function(lookupControlle
 
         this.addAppServicePath({
             method: "GET",
-            path: "/_matrix/app/:version/thirdparty/protocol/:protocol",
+            path: "/_matrix/app/:version(v1|unstable)/thirdparty/protocol/:protocol",
             checkToken: this.opts.authenticateThirdpartyEndpoints,
             handler: function(req, res) {
                 const protocol = req.params.protocol;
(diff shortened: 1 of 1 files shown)


src/bridge.js
```diff
@@ -445,7 +445,7 @@ Bridge.prototype._customiseAppserviceThirdPartyLookup = function(lookupControlle
 
         this.addAppServicePath({
             method: "GET",
-            path: "/_matrix/app/:version/thirdparty/protocol/:protocol",
+            path: "/_matrix/app/:version(v1|unstable)/thirdparty/protocol/:protocol",
```
jaller94, line 448: `/:version(v1|unstable)/` is the same as `/(v1|unstable)/`?
It does not allow URLs like `/invalidv1/` where `version="invalid"`?

jaller94: Yep, these format restrictions for route params are described over here:
https://expressjs.com/en/guide/routing.html#route-parameters

Half-Shot: image

Try `/_matrix/app/(v1|unstable)/thirdparty/protocol/:protocol`
and `/_matrix/app/:version(v1|unstable)/thirdparty/protocol/:protocol`
against the test string `/_matrix/app/v1/thirdparty/protocol/p`

http://forbeslindesay.github.io/express-route-tester/

Apparently if the group is not named, it doesn't work.

Half-Shot: > It does not allow URLs like /invalidv1/ where version="invalid"?

Oh and yes, correct. Just like a RegExp

PR #162, opened 2020-04-17
Header is lowercase

Half-Shot:
Missed this during testing. `express` lowercases header names :/


jaller94:
Any idea why `headers` is not in the API Reference?
https://expressjs.com/en/api.html#req

Half-Shot:
Presumably I should have done `req.get` rather than `req.headers` :facepalm:

PR #166, opened 2020-05-17
Port prometheus components to TypeScript

Half-Shot:
Started out as a mission to fix the deprecation warnings caused by metrics, I ended up porting the whole thing.


src/components/prometheusmetrics.ts
```diff
@@ -0,0 +1,394 @@
...
+            help: "Count of the number of Matrix client API calls made",
+            labels: ["method"],
+        });
+
+        /*
+            * We'll now annotate a bunch of the methods in MatrixClient to keep counts
+            * of every time they're called. This seems to be neater than trying to
+            * intercept all HTTP requests and try to intuit what internal method was
+            * invoked based on the HTTP URL.
+            * It's kindof messy to do this because we have to maintain a list of
```
jaller94, line 129:
```suggestion
            * It's kind of messy to do this because we have to maintain a list of
```

src/index.ts
```diff
@@ -48,8 +48,9 @@ module.exports.ContentRepo = (
 	require("matrix-js-sdk").ContentRepo
 );
 
-module.exports.PrometheusMetrics = require("./components/prometheusmetrics");
-module.exports.PrometheusMetrics.AgeCounters = require("./components/agecounters");
+
+export * from "./components/prometheusmetrics";
+module.exports.PrometheusMetrics.AgeCounters = require("./components/agecounters").AgeCounters;
```
jaller94, line 53: That's ok without defining `module.exports.PrometheusMetrics` explicitly??
ok, i guess..

Half-Shot: Ah, it's the above export line. It's sugar syntax for `module.exports.PrometheusMetrics = require("./components/prometheusmetrics");`

jaller94: Just feels super weird to patch an object that has not been defined explicitly and hope the export name in the other file does not change.

PR #170, opened 2020-06-25
Fix prototype bug from the prometheus PR

Half-Shot:


src/components/prometheusmetrics.ts
```diff
@@ -135,7 +135,7 @@ export class PrometheusMetrics {
             * the lesser of two evils.
             */
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        const matrixClientPrototype = (MatrixClient as any).prototype;
+        const matrixClientPrototype = (JsSdk as any).MatrixClient.prototype;
```
jaller94, line 138: Is `as any` still required because `matrix-js-sdk` does not ship with types?

src/components/prometheusmetrics.ts
```diff
@@ -15,7 +15,7 @@ limitations under the License.
 
 import PromClient, { Registry } from "prom-client";
 import { AgeCounters } from "./agecounters";
-import MatrixClient from "matrix-js-sdk";
+import JsSdk from "matrix-js-sdk";
```
jaller94, line 18: Can you `{ MatrixClient }` if you don't need other stuff from the Sdk?

src/components/prometheusmetrics.ts
Half-Shot: It doesn't ship with MatrixClient exported as a type, despite being there. This is a ugly hack, so it causes uglyness :(

src/components/prometheusmetrics.ts
Half-Shot: Sadly not, as it's not exported as a type. You can access it, but TS will complain.

PR #175, opened 2020-07-24
Update packages

Half-Shot:
Includes some minor linting fixes as a result of the eslint bump


Half-Shot:
(I've removed travis from the repo settings, we can ignore those failures)

jaller94 (approved):
I find the Bluebird and chalk upgrades the most worrying. But lets get it out and test it in the wild.
A quick look through their changelogs didn't bring up anything that should break for us..

PR #177, opened 2020-07-24
Typescriptify event queue

Half-Shot:
This is a fairly easy candidate to port, as it was using ES6 classes already.

### DIFF
--- a/src/components/event-queue.ts
+++ b/src/components/event-queue.ts
@@ -12,8 +12,10 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
+import Bluebird from "bluebird";
 
-const Bluebird = require("bluebird");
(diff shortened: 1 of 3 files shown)


jaller94:
Please fix the linting errors first.

src/components/event-queue.ts
```diff
@@ -33,15 +38,12 @@ class EventQueue {
...
-            // $identifier: {
-            //  events: [ {dataReady: } ],
-            //  consuming: true|false
-            // }
-        };
-        this.consumeFn = consumeFn;
+    private queues: { [identifer: string]: {
+        events: Array<{ dataReady: DataReady }>;
+        consuming: boolean;
+    }; } = {};
```
jaller94, line 44: This line and the two above are allowed to have semicolons??

Half-Shot: Yeah, it's a substitute for a comma. The linter informs me that was the style choice we went for :thinking:

PR #178, opened 2020-07-24
Run build on local install

Half-Shot:
This runs the build step before publishing AND when the repository is installed standalone, which will fix the test failures.


jaller94 (approved):
ok

PR #179, opened 2020-07-24
Fix usages of private properties on AppserviceRegistration

Half-Shot:


Half-Shot:
It did appear to work from my testing :)

PR #184, opened 2020-08-03
Convert RoomLinkValidator to TypeScript and fix errors

### LINKED ISSUE #183: RoomLinkValidator terminates validation on the first ignored user
Instead of `break`, this should have been a `continue` to skip the current `userId` but not all following.

https://github.com/matrix-org/matrix-appservice-bridge/blob/d2467d5ed56540b18d508dd7ed005bc7ef3582df/src/components/room-link-validator.js#L136

jaller94:
* ⚠️ Declares some properties and methods private that could have been considered public. As such this is potentially API-breaking.
* Fixes #183.


src/components/room-link-validator.js
```diff
@@ -1,175 +0,0 @@
...
-    /**
-     * @param {string} config Config for the validator.
-     * @param {string} config.ruleFile Filename for the rule file.
-     * @param {string} config.rules Rules if not using a rule file, will be
-     *                              overwritten if both is set.
-     * @param {AppServiceBot} asBot The AS bot.
-     */
-    constructor (config, asBot) {
-        this.conflictCache = new Map(); // roomId => number
-        this.waitingRooms = new Map(); // roomId => Promise
```
jaller94, line 67: I guess this was meant to stop the same room from getting processed twice at the same time, but it wasn't implemented.

jaller94:
I don't understand how to get a property from `unknown` when I have already proven that it is an `object`.
Screenshot_2020-08-04_13-46-58

PR #185, opened 2020-08-03
Convert intent.js to TypeScript

jaller94:
Breaking change:
* It's no longer ok to hand `new Intent(client, client, opts)` to Intent and modify `opts`. `Intent` will now create a deep clone for itself and ignore outside changes.


Half-Shot:
Hah, I wrote a branch for this on Friday but never published it. Ah well. This looks very clean.

src/components/intent.ts
```diff
@@ -0,0 +1,829 @@
...
+
+    /**
+     * <p>Send a plaintext message to a room.</p>
+     * This will automatically make the client join the room so they can send the
+     * message if they are not already joined. It will also make sure that the client
+     * has sufficient power level to do this.
+     * @param roomId The room to send to.
+     * @param text The text string to send.
+     */
+    sendText = (roomId: string, text: string) => {
```
Half-Shot, line 237: Could these be `public sendText` methods?

jaller94: Are they not public by default?

Half-Shot: Yes but it looks very ugly to use the  `func = (args)`. Can we drop the `=`. I'm leaning towards including `public` for readability, but I'm not so bothered about it.

jaller94: Yeah, we can drop `func = (args) => {}`. I usually use arrow functions whenever I can, because I don't trust `this` in `function() {}` anymore.

Half-Shot: You can trust class methods :). They are here for you.

Half-Shot:
@jaller94 do you need any help with these tests?

jaller94:
Haven't had time to fix them. I assume
* I messed up some condition OR
* type checks became stricter and the test are providing some insufficient mocks.

Half-Shot:
Might be worth factoring in https://github.com/matrix-org/matrix-appservice-bridge/pull/155, if you've not already?

src/components/intent.ts
```diff
@@ -0,0 +1,850 @@
...
+    * @param opts.dontJoin True to not attempt to join a room before
+    * sending messages into it. The surrounding code will have to ensure the correct
+    * membership state itself in this case. Default: false.
+    *
+    * @param opts.enablePresence True to send presence, false to no-op.
+    *
+    * @param opts.caching.ttl How long requests can stay in the cache, in milliseconds.
+    * @param opts.caching.size How many entries should be kept in the cache, before the oldest is dropped.
+    */
+    constructor(private client: MatrixClient, private botClient: MatrixClient, opts: IntentOpts = {}) {
```
Half-Shot, line 166: client shouldn't be private, it's deliberately exposed for bridges to use if they need access to the js-sdk

src/components/intent.ts
```diff
@@ -0,0 +1,850 @@
...
+*/
+
+import MatrixUser from "../models/users/matrix";
+import JsSdk from "matrix-js-sdk";
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const { MatrixEvent, RoomMember } = JsSdk as any;
+import ClientRequestCache from "./client-request-cache";
+import { defer } from "../utils/promiseutil";
+
+type MatrixClient = {
```
Half-Shot, line 23: My concern with this is what we should probably be using matrix-js-sdk's types. Although I believe they do not exist for the time being, I am worried by strictly typing us to just these functions, anything extra that the bridges may be using (for example, uploadContent) won't be accessible without type hacks.

I am strongly against us doing types for matrix-js-sdk, as they often have breaking releases.

src/components/intent.ts
```diff
@@ -0,0 +1,850 @@
...
+
+    /**
+     * <p>Join a room</p>
+     * This will automatically send an invite from the bot if it is an invite-only
+     * room, which may make the bot attempt to join the room if it isn't already.
+     * @param roomId The room to join.
+     * @param viaServers The server names to try and join through in
+     * addition to those that are automatically chosen.
+     */
+    public async join(roomId: string, viaServers: string[]) {
```
Half-Shot, line 514:
```suggestion
    public async join(roomId: string, viaServers?: string[]) {
```

src/components/intent.ts
```diff
@@ -0,0 +1,850 @@
...
+    /**
+     * <p>Kick a user from a room.</p>
+     * This will automatically make the client join the room so they can send the
+     * kick if they are not already joined.
+     * @param roomId The room to kick the user from.
+     * @param target The target of the kick operation.
+     * @param reason Optional. The reason for the kick.
+     * @return Resolved when kickked, else rejected with an error.
+     */
+    public async kick(roomId: string, target: string, reason: string) {
```
Half-Shot, line 474:
```suggestion
    public async kick(roomId: string, target: string, reason?: string) {
```

src/components/intent.ts
```diff
@@ -0,0 +1,850 @@
...
+    /**
+     * <p>Ban a user from a room.</p>
+     * This will automatically make the client join the room so they can send the
+     * ban if they are not already joined.
+     * @param roomId The room to ban the user from.
+     * @param target The target of the ban operation.
+     * @param reason Optional. The reason for the ban.
+     * @return Resolved when banned, else rejected with an error.
+     */
+    public async ban(roomId: string, target: string, reason: string) {
```
Half-Shot, line 488:
```suggestion
    public async ban(roomId: string, target: string, reason?: string) {
```

src/components/intent.ts
jaller94: `matrix-js-sdk` doesn't currently provide types. These are mainly used internally, but also dictate the return type of any method returning the. 
A lot of methods of intent are just relaying promises from the client.
An alternative would be to import `matrix-js-sdk` (implicitly as `any`) and specifying the returns types explicitly.
Either way, as long as matrix-js-sdk isn't providing types, this file will need to specify some for it.

Yes, the current implementation gives false certainty over functions, parameters and return types that may or may not exist.

Half-Shot: I think it would be better to just keep it as `any` for now, and wait for types to exist. We can sugar coat it somewhat by having explicit return types on the intent functions.

Half-Shot:
Heads up that I've been trying to incorporate types into matrix-appservice-bridge based off this, so I'll have a PR shortly with what I changed to make it work.

src/components/intent.ts
```diff
@@ -0,0 +1,850 @@
...
+                event_id: "_",
+                state_key: "",
+                type: "m.room.power_levels"
+            }
+            const powerLevelEvent = new MatrixEvent(event);
+            // What level do we need for this event type?
+            const defaultLevel = STATE_EVENT_TYPES.includes(eventType)
+                ? event.content.state_default
+                : event.content.events_default;
+            const requiredLevel = event.content.events[eventType] || defaultLevel;
```
jaller94, line 794: If the powerLevelEvent's content did not contain a required level for the eventType and also no defaults, what should happen? Do we just assume the required power level is higher than what we have?

Half-Shot: I believe in theory synapse ensures that should never happen, but we *should* [follow the spec](https://matrix.org/docs/spec/client_server/r0.6.0#m-room-power-levels) which mentions what values to use if the keys do not exist.

PR #187, opened 2020-08-06
Allow linter warnings but be strict on rules

Half-Shot:
So we can have greater visibility on what needs fixing, rather than resorting to ignore comments.


.eslintrcts.json
```diff
@@ -13,72 +13,73 @@
...
+    "no-spaced-func": "error",
+    "no-trailing-spaces": "error",
+    "keyword-spacing": ["error", {"before": true, "after": true}],
+    "space-before-blocks": ["error", "always"],
+    "@typescript-eslint/explicit-function-return-type": "off",
     "camelcase": ["error", { "properties": "never" }],
-    "@typescript-eslint/ban-ts-ignore": 0,
-    "no-unused-vars": 0,
+    "@typescript-eslint/ban-ts-ignore": "off",
+    "no-unused-vars": "off",
```
jaller94, line 79: 🤨

Half-Shot: ?

jaller94: I am wondering why unused variables need to be allowed for now, but I don't think it's a blocker.

Half-Shot: Oh, it was before. I just changed it to a more readable string.

PR #188, opened 2020-08-07
Use own implementation of defer

### LINKED ISSUE #180: We should migrate off deferreds
http://bluebirdjs.com/docs/api/deferred-migration.html has been around since 2018, and it's now quite spammy (`Promise.defer is deprecated and will be removed in a future version. Use new Promise instead.`). Ideally, let's chuck out Bluebird.

Half-Shot:
Fixes #180 

Stops the warnings. This defer implementation was copied from matrix-appservice-irc.

### DIFF
--- a/changelog.d/188.bugfix
+++ b/changelog.d/188.bugfix
@@ -1,0 +1,1 @@
+Fix bluebird defer warnings by using our own defer implementation.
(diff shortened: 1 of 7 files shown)


jaller94 (approved):
Looks good to me.

PR #189, opened 2020-08-07
Port Request(Factory) to Typescript

Half-Shot:
Just another easy one to do.


src/components/request.ts
```diff
@@ -0,0 +1,113 @@
...
+     * Resolve or reject the promise depending on the outcome of this promise.
+     * @param promise The promise whose resolution determines the outcome of this
+     * request.
+     */
+    public outcomeFrom(promise: Promise<unknown>) {
+        return promise.then(
+            (msg) => this.resolve(msg)
+        ).catch(
+            (msg) => this.reject(msg)
+        );
```
jaller94, line 111: Should be the same as this:
```suggestion
        return promise.then(this.resolve, this.reject);
```

Half-Shot: True.

PR #191, opened 2020-08-07
Await for a Bridge to listen to its port

### LINKED ISSUE #192: Bridge.run() spawns a floating promise for listening to a port
https://github.com/matrix-org/matrix-appservice-bridge/blob/8c85b240441b742f191707098e74d8a10b1ed1a8/src/bridge.js#L377

We should await this function as it returns a promise that we may need to handle.

jaller94:
Fixes #192 

This allows for errors of `.listen` to bubble up and get handled.

### DIFF
--- a/src/bridge.js
+++ b/src/bridge.js
@@ -376,11 +376,10 @@ Bridge.prototype.run = function(port, config, appServiceInstance, hostname) {
     }
 
     // We MUST return a Bluebird-Promise instead of a Promise.
-    // This is needed by many tests in this repo.
-    return new Bluebird(async() => {
+    // promise.done() is used by many tests in this repo.
+    return this.loadDatabases().then(async() => {
         await this.appService.listen(port, hostname);
-        return this.loadDatabases();
-    })
+    });
 };
 
 /**
--- a/src/bridge.js
+++ b/src/bridge.js
@@ -248,7 +248,7 @@ function Bridge(opts) {
 
 /**
  * Load the user and room databases. Access them via getUserStore() and getRoomStore().
- * @return {Promise} Resolved/rejected when the user/room databases have been loaded.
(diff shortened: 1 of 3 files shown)


jaller94:
I will need to return a Bluebird promise for our tests.
Also, this may have increased the time we have to wait beyond Jasmine's default timeout.

Half-Shot:
How come? This shouldn't affect the tests?

jaller94:
Some tests use `returnValue.done()` instead of `.then()` as the last handler, which only works if `.run()` returns a Bluebird-Promise, but not a native Promise.

Half-Shot:
oh, fair enough then.

src/bridge.js
```diff
@@ -374,8 +375,11 @@ Bridge.prototype.run = function(port, config, appServiceInstance, hostname) {
         this._metrics.addAppServicePath(this);
     }
 
-    this.appService.listen(port, hostname);
-    return this.loadDatabases();
+    // We MUST return a Bluebird-Promise instead of a Promise.
+    // promise.done() is used by many tests in this repo.
+    return this.loadDatabases().then(async() => {
```
Half-Shot, line 380: Stupid question, why don't we use await here?

jaller94: Great question.
`loadDatabase` must be executed immediately to initialize the stores needed by some promises above. These promises will run as soon as `run` is done. The only other way is to move `loadDatabase` above the usage of `Promise.cast()`&`this._onAliasQuery()` and await it there.

`loadDatabase` returns a Bluebird-Promise, which is exactly what I need to return to make the tests happy. I just chain `.listen` in a native promise at the end, but the return value will stay a Bluebird-Promise.

Half-Shot: Fair enough, what a mess though :S

PR #193, opened 2020-08-07
Fix usage of deferPromise

jaller94:
In the function loadDatabase we use `const defer = defer()`. This means the second `defer` is not the imported function, but the undefined variable of the local scope.


jaller94:
Removed the label bug as this hadn't been released.

jaller94:
? CI is cool with me not adding a newsfile?

Half-Shot:
>? CI is cool with me not adding a newsfile?

surprising

PR #195, opened 2020-08-10
Convert Cli to Typescript

Half-Shot:
Had some fun here with generics :). Generally feels like this is going to be useful for people hacking on bridges.


src/components/cli.ts
```diff
@@ -0,0 +1,296 @@
...
+     * can be accessed via <code>Cli.getConfig()</code>.
+     * @param {string|Object=} opts.bridgeConfig.schema Path to a schema YAML file
+     * (string) or the parsed schema file (Object).
+     * @param {Object=} opts.bridgeConfig.defaults The default options for the
+     * config file.
+     * @param {boolean=} opts.enableRegistration Enable '--generate-registration'.
+     * Default True.
+     * @param {string=} opts.registrationPath The path to write the registration
+     * file to. Users can overwrite this with -f.
+     * @param {boolean=} opts.enableLocalpart Enable '--localpart [-l]'. Default: false.
```
jaller94, line 76: Can we remove the JSDoc types then?

jaller94, line 76:
```suggestion
     * @param opts CLI options
     * @param opts.run The function called when you should run the bridge.
     * @param opts.generateRegistration The function
     * called when you should generate a registration.
     * @param opts.bridgeConfig Bridge-specific config info. If null, no
     * --config option will be present in the CLI. Default: null.
     * @param opts.bridgeConfig.affectsRegistration True to make the
     * --config option required when generating the registration. The parsed config
     * can be accessed via <code>Cli.getConfig()</code>.
     * @param opts.bridgeConfig.schema Path to a schema YAML file
     * (string) or the parsed schema file (Object).
     * @param opts.bridgeConfig.defaults The default options for the
     * config file.
     * @param opts.enableRegistration Enable '--generate-registration'.
     * Default True.
     * @param opts.registrationPath The path to write the registration
     * file to. Users can overwrite this with -f.
     * @param opts.enableLocalpart Enable '--localpart [-l]'. Default: false.
```

PR #196, opened 2020-08-10
Type fixes needed to make matrix-appservice-irc build

Half-Shot:
This is a collection of changes needed to make our types work with the IRC bridge. We should either merge this into the parent PR (and subsequently into that one), or find a way to include these changes.

### DIFF
--- a/src/components/state-lookup.ts
+++ b/src/components/state-lookup.ts
@@ -18,7 +18,7 @@ import PQueue from "p-queue";
 interface StateLookupOpts {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     client: any; //TODO: Needs to be MatrixClient (once that becomes TypeScript)
-    stateLookupConcurrency: number;
+    stateLookupConcurrency?: number;
     eventTypes?: string[];
     retryStateInMs?: number;
 }
@@ -33,7 +33,7 @@ interface StateLookupRoom {
     };
 }
 
-interface StateLookupEvent {
+export interface StateLookupEvent {
     // eslint-disable-next-line camelcase
     room_id: string;
     // eslint-disable-next-line camelcase
@@ -41,6 +41,7 @@ interface StateLookupEvent {
     type: string;
     // eslint-disable-next-line camelcase
     event_id: string;
+    content: unknown;
 }
 
 const RETRY_STATE_IN_MS = 300;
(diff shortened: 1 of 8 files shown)


src/components/state-lookup.ts
```diff
@@ -33,14 +33,15 @@ interface StateLookupRoom {
...
-interface StateLookupEvent {
+export interface StateLookupEvent {
     // eslint-disable-next-line camelcase
     room_id: string;
     // eslint-disable-next-line camelcase
     state_key: string;
     type: string;
     // eslint-disable-next-line camelcase
     event_id: string;
+    content: unknown;
```
jaller94, line 44: Is content guaranteed to be an `Object` or claimed as here `unknown`? That would require some more type checks in my PR.

Half-Shot: We could probably call it a `Record<string,unknown>`

jaller94: But if that's not guaranteed by the client lib, we'll get errors like the one #155 fixes, where a presumed object may be a string or a number.

Half-Shot: Content should really never be anything other than a object, if it is, we should reject it before insertion?

Half-Shot: https://github.com/matrix-org/matrix-appservice-bridge/pull/196/commits/e027f071835160d106616950564bf8d787d3531f

src/components/state-lookup.ts
```diff
@@ -191,9 +187,22 @@ export class StateLookup {
...
-            r.events[event.type] = {};
+        this.insertEvent(r, event);
+    }
+
+    private insertEvent(roomSet: StateLookupRoom, event: StateLookupEvent) {
+        if (typeof event.content !== "object") {
+            // Reject - unexpected content type
+            return;
+        }
+        if (!event.type || !event.state_key) {
```
jaller94, line 198: Just for learning something: This broke the tests. It's not testing if these keys are missing (as claimed by the line below), but if they are truethy. The tests use empty strings, which aren't truethy.

jaller94: Fixed in https://github.com/matrix-org/matrix-appservice-bridge/pull/185/commits/deb2f38ab33b74567bba6be879440da1d26f5a4b

PR #199, opened 2020-08-11
Use Typedoc over JSDoc

### LINKED ISSUE #198: Docs cannot process TypeScript
We need to find a way to either export jsdoc format docs onto our compiled js, or perhaps just use http://typedoc.org

Half-Shot:
Fixes #198 

Example output:

Screenshot from 2020-08-11 11-48-28


jaller94 (approved):
Most of our code is TypeScript no anyway.
Typedoc looked quite ok locally. :)

PR #200, opened 2020-08-11
minor type fixes

Half-Shot:


src/components/cli.ts
```diff
@@ -15,7 +15,7 @@ limitations under the License.
 */
 
 import * as fs from "fs";
-import * as path from "path";
+import path from "path";
```
jaller94, line 18: `path` gets imported like this, but `fs` above does not?

Half-Shot: apparently we needed the default path but not for fs

PR #202, opened 2020-08-17
Typescriptify models/*

Half-Shot:


src/models/users/remote.ts
```diff
@@ -0,0 +1,64 @@
...
+        return this.data[key] as T;
+    }
+
+    /**
+     * Set an arbitrary bridge-specific data value for this room.
+     * @param key The key to store the data value under.
+     * @param val The data value. This value should be serializable via
+     * <code>JSON.stringify(data)</code>.
+     */
+    public set<T>(key: string, val: T) {
```
jaller94, line 61: I don't see the value in declaring T.
```suggestion
    public set(key: string, val: unknown) {
```

src/models/users/remote.ts
```diff
@@ -0,0 +1,64 @@
...
+        return this.data;
+    }
+
+    /**
+     * Get the data value for the given key.
+     * @param key An arbitrary bridge-specific key.
+     * @return Stored data for this key. May be undefined.
+     */
+    public get<T>(key: string) {
+        return this.data[key] as T;
```
jaller94, line 52: Basically endless possibilities to shoot yourself in the foot by declaring an incorrect type.
I think the return value should be `unknown`. The function calling this may cast an `as T` if it's so sure about the return type.

src/models/users/remote.ts
```diff
@@ -0,0 +1,64 @@
...
+*/
+
+export class RemoteUser {
+
+    /**
+     * @param identifier The unique ID for this user.
+     * @param data The serialized key-value data object to assign to this user.
+     * @throws If identifier is not supplied.
+     */
+    constructor (public readonly id: string, public data: Record<string, unknown>) {
```
jaller94, line 23: Why is data public when there's also `get` and `set`?

jaller94, line 23: Why is data not readonly? This allows calls from the the outside to redefine it.

src/models/users/matrix.ts
```diff
@@ -0,0 +1,126 @@
...
+     *               to change the default value.
+     */
+    constructor(public userId: string, public readonly _data: Record<string, unknown> = {}, escape=MatrixUser.ESCAPE_DEFAULT) {
+        if (!userId) {
+            throw Error("Missing user_id");
+        }
+        if (_data && typeof _data !== "object") {
+            throw Error("data arg must be an Object");
+        }
+        this.userId = userId;
```
jaller94, line 34: Isn't that redundant to the constructor's definition of userId being public?

src/models/users/matrix.ts
```diff
@@ -0,0 +1,126 @@
...
+    public readonly host: string;
+
+    /**
+     * Construct a Matrix user.
+     * @param userId The userId of the user.
+     * @param data Serialized data values
+     * @param escape Escape the user's localpart. Modify {@link MatrixUser~ESCAPE_DEFAULT}
+     *               to change the default value.
+     */
+    constructor(public userId: string, public readonly _data: Record<string, unknown> = {}, escape=MatrixUser.ESCAPE_DEFAULT) {
```
jaller94, line 27: The leading `_` suggests that it was always meant to be private.
```suggestion
    constructor(public userId: string, private readonly _data: Record<string, unknown> = {}, escape=MatrixUser.ESCAPE_DEFAULT) {
```

jaller94 (approved):
I think that the set and get functions should use `unknown` instead of `T` which they don't verify.
It also worries me that a lot of JS objects are copied by reference instead of deep cloning or deep freezing the content to prevent further modification from the outside.

But since it's an improvement over the last version and my suggestions from above may introduce API-breaking changes (by preventing unsafe behaviours), I'm approving this PR.

src/models/users/remote.ts
Half-Shot: Yeah, I think my reason was because the first thing you'll do here is use `get("foo") as string` in the bridge code because undefined is pretty useless. As you say though, it will lead to undefined behavior :/

src/models/users/remote.ts
Half-Shot: I think in reality some bits of bridge code use `data` and since it wasn't underlined I didn't want to break code.

PR #207, opened 2020-08-23
Add ability to watch config for changes

Half-Shot:
This allows bridge developers to listen for config changes, which can be bubbled up to the bridge logic to support small changes without restarting the whole bridge. This is useful for things like the irc bridge, where restarts are costly.

I've also added some minimal Cli tests to ensure that config behaviour continues to work as expected.

### DIFF
--- a/src/components/cli.ts
+++ b/src/components/cli.ts
@@ -249,7 +249,7 @@ export class Cli<ConfigType extends Record<string, unknown>> {
             fs.watchFile(this.args.config, {
                 persistent: false,
                 interval: this.opts.bridgeConfig.watchInterval || Cli.DEFAULT_WATCH_INTERVAL },
-                (curr: fs.Stats) => {
+                () => {
                 log.info("Config file change detected, reloading");
                 try {
                     const newConfig = this.loadConfig(this.args?.config);
@@ -257,7 +257,8 @@ export class Cli<ConfigType extends Record<string, unknown>> {
                     if (newConfig && this.opts.onConfigChanged) {
                         this.opts.onConfigChanged(newConfig);
                     }
-                } catch (ex) {
+                }
+                catch (ex) {
                     log.warn("Failed to reload config file:", ex);
                 }
             });
--- a/src/components/cli.ts
+++ b/src/components/cli.ts
@@ -195,7 +195,7 @@ export class Cli<ConfigType extends Record<string, unknown>> {
     private assignConfigFile(configFilePath: string) {
         const configFile = (this.opts.bridgeConfig && configFilePath) ? configFilePath : undefined;
         if (!configFile) {
-            return null;
+            return;
         }
         const config = this.loadConfig(configFile);
         this.bridgeConfig = config;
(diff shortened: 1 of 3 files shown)


dalcde:
I was thinking of a different workflow here, but I don't have strong opinions. The way, say, `apache` and `nginx` do this is that after you change the config, you send a `SIGHUP` signal to the process, instead of modifying the file directly.

This would interact appropriately with `systemctl reload` and can let people decide when they want to actually reload the config file (e.g. they might want to regenerate the registration file or do something else first). It's also helpful for people who want to edit config on a live instance (e.g. on a staging server) and want to be able to save it halfway through edits because of muscle memory.

Half-Shot:
@dalcde Great idea. I'll definitely update the code to use signals.

Half-Shot:
This has been done.

src/components/cli.ts
```diff
@@ -229,7 +233,24 @@ export class Cli<ConfigType extends Record<string, unknown>> {
...
+        if (this.opts.onConfigChanged && this.opts.bridgeConfig) {
+            log.info("Will listen for SIGHUP");
+            process.on("SIGHUP",
+                () => {
+                log.info("Got SIGHUP, reloading config file");
+                try {
+                    const newConfig = this.loadConfig(configFilename);
+                    // onConfigChanged is checked above, but for Typescript's sake.
+                    if (this.opts.onConfigChanged) {
+                        this.opts.onConfigChanged(newConfig);
```
dalcde, line 246: Do you want to update the `this.bridgeConfig` variable here?

In real life, only some config variables would be hot-reloadable. I would expect that if the new config edits other variables, the bridge should reject the new config file and retain the old one. We can make `onConfigChanged` return a boolean to indicate whether the new config file is accepted, and act based on that.

Half-Shot: Nope, I don't want to update the config from underneath bridges because it could cause problems. For instance, certain bits of the IRC bridge cannot be reloaded so we wouldn't want to change those variables.

It will be up to the bridge developer to listen for the callback, and explicitly apply changes to the config inside the bridge. It might be a bit more faff, but it's safer.

dalcde:
One issue I envsion is that bridgeConfig is a read-only variable, so it
is impossible to reassign it to the new config.

src/components/cli.ts
```diff
@@ -229,7 +233,24 @@ export class Cli<ConfigType extends Record<string, unknown>> {
...
+    private startWithConfig(config: ConfigType|null, configFilename: string) {
+        if (this.opts.onConfigChanged && this.opts.bridgeConfig) {
+            log.info("Will listen for SIGHUP");
+            process.on("SIGHUP",
+                () => {
+                log.info("Got SIGHUP, reloading config file");
+                try {
+                    const newConfig = this.loadConfig(configFilename);
+                    // onConfigChanged is checked above, but for Typescript's sake.
+                    if (this.opts.onConfigChanged) {
```
jaller94, line 245: Not just for TypeScript's sake. We're copying the `opts` object by reference.
Any code in this class or the bridge could have re-set the value of `opts.onConfigChanged` before `SIGHUP` was received.

jaller94 (approved):
Nice. Also great to see tests covering this reload functionality.

Off-topic:
I wonder if we should deep-copy or deep-freeze the `opts` parameter, because it seems like you expect it to be immutable.

PR #211, opened 2020-08-23
Port errors.js to Typescript

Half-Shot:


src/errors.ts
```diff
@@ -0,0 +1,95 @@
...
+limitations under the License.
+*/
+
+export namespace unstable {
+    /**
+     * Append the old error message to the new one and keep its stack trace.
+     * Example:
+     *     throw wrapError(e, HighLevelError, "This error is more specific");
+     */
+    export function wrapError<T extends Error>(
```
jaller94, line 22: Do we have to flag this renamed function as a breaking change or can it not be used by external software?

jaller94 (approved):
Looks fine to me.
If `wrap` was public, I would like to mention this as a deprecation in the changelog, and have an alias for a while.

src/errors.ts
Half-Shot: This release is going to be breaking (2.0) so I'm not concerned about an alias, but we should call out that it's changing name.

jaller94: bth. I would be annoyed by a lib making this seemingly trivial name change.

It would be a bit nicer with some grace period where the old function is an alias and console log's that it's going to be deprecated on first usage.

jaller94: If the parent project doesn't use TypeScript this will likely fail on runtime. And, given this is related to error logging, it may become though to see if a component or the error reporting failed.

Half-Shot: I'll add a warning.

jaller94 (approved):
Perfect :)

PR #215, opened 2020-08-29
Don't ignoreCache in Intent.join

dalcde:
This fixes a race condition where joining a room immediately after leaving it does not actually result in a join.

This is consistently reproducable on my setup, where this change fixes the issue. However, extracting the offending code is probably difficult. I believe the cause is that `Intent.leave` acts via the client, which doesn't update the cache.

### DIFF
--- a/src/components/intent.ts
+++ b/src/components/intent.ts
@@ -494,7 +494,7 @@ export class Intent {
      * addition to those that are automatically chosen.
      */
     public async join(roomId: string, viaServers?: string[]) {
-        await this._ensureJoined(roomId, false, viaServers);
+        await this._ensureJoined(roomId, true, viaServers);
     }
 
     /**


src/components/intent.ts
```diff
@@ -494,7 +494,7 @@ export class Intent {
      * addition to those that are automatically chosen.
      */
     public async join(roomId: string, viaServers?: string[]) {
-        await this._ensureJoined(roomId, false, viaServers);
+        await this._ensureJoined(roomId, true, viaServers);
```
Half-Shot, line 497: I think this might be the wrong solution. The cache is useful as bridges like IRC will often do joins multiple times for users if it's missing state, as fundamentally bridges are usually not expected to cache state.

I'd be happy if we used `this.opts.backingStore.setMembership` `this.opts.backingStore.getMembership` in `leave` though.

dalcde:
I converted this to an issue at #229. I don't think I would be able to
work on this in the near future.

PR #216, opened 2020-09-01
Remove bluebird dependency

Half-Shot:
This is a breaking change to enforce the use of **pure** Bluebird promises. Since we've already made breaking changes in other areas of the library, we might as well remove the use of these as well.

### DIFF
--- a/src/components/state-lookup.ts
+++ b/src/components/state-lookup.ts
@@ -153,8 +153,8 @@ export class StateLookup {
         }
         r.events = {};
 
+        r.syncPending = true;
         r.syncPromise = (async () => {
-            r.syncPending = true;
             const res = await this.getInitialState(roomId);
             r.syncPending = false;
             return res;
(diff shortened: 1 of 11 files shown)


src/components/state-lookup.ts
```diff
@@ -146,13 +145,20 @@ export class StateLookup {
...
-        r.syncPromise = Bluebird.resolve(this.getInitialState(roomId));
+
+        r.syncPromise = (async () => {
+            r.syncPending = true;
+            const res = await this.getInitialState(roomId);
+            r.syncPending = false;
+            return res;
+        })();
 
         return r.syncPromise;
```
jaller94, line 163: We may want to ensure the state is set to pending before the promise gets run as we already know its pending.
```suggestion
        r.syncPending = true;
        r.syncPromise = (async () => {
            const res = await this.getInitialState(roomId);
            r.syncPending = false;
            return res;
        })();

        return r.syncPromise;
```

jaller94 (approved):
Looking good

PR #217, opened 2020-09-02
Make url parameter optional when generating registration

### LINKED ISSUE #182: Allow url parameter to be optional
Currently, if we pass the `-r` flag to the cli to generate a registration, it forces the user to supply the url (`-u`). However, this is not used if the bridge chooses to let the user configure the url via the config file instead. There should be an option when constructing the `Cli` to disable the `-u` flag.

https://github.com/matrix-org/matrix-appservice-bridge/blob/develop/src/components/cli.js#L121

dalcde:
The bridge may instead choose to let the user configure the url.

Fixes #182

### DIFF
--- a/changelog.d/217.feature
+++ b/changelog.d/217.feature
@@ -1,0 +1,1 @@
+Make url parameter optional when generating registration.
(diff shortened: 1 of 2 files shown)


Half-Shot (approved):
Nice and concise. Happy with this.

PR #218, opened 2020-09-03
Various type improvements to improve compatibility with existing bridges

Half-Shot:
This PR is a mixed bag of small code changes to ensure compatibility with existing bridges. Care has been taken to lean towards updating the IRC bridge where possible, rather than porting bad behaviours back into this SDK.

### DIFF
--- a/src/components/room-bridge-store.ts
+++ b/src/components/room-bridge-store.ts
@@ -274,7 +274,7 @@ export class RoomBridgeStore extends BridgeStore {
                 return e.remote;
             }) as RemoteRoom[];
         }
-        return entryMap;
+        return result;
     }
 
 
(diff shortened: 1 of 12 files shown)


src/components/room-bridge-store.ts
```diff
@@ -274,7 +274,7 @@ export class RoomBridgeStore extends BridgeStore {
                 return e.remote;
             }) as RemoteRoom[];
         }
-        return entryMap;
+        return result;
```
jaller94, line 277: :D Nice catch.
(For how long has this been broken? D: )

jaller94 (approved):
Looks good to me.

src/components/room-bridge-store.ts
Half-Shot: Only for the Typescript bits, luckily.

PR #221, opened 2020-09-04
Import @types of dependencies

jaller94:
According to the TypeScript docs, we should also make the `@types` of our dependencies direct dependencies.

https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#dependencies

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,10 @@
   },
   "homepage": "https://github.com/matrix-org/matrix-appservice-bridge#readme",
   "dependencies": {
+    "@types/extend": "^3.0.1",
+    "@types/js-yaml": "^3.12.5",
+    "@types/nedb": "^1.8.10",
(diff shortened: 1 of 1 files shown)


Half-Shot:
>Our package exposes declarations from each of those, so any user of our browserify-typescript-extension package needs to have these dependencies as well. For that reason, we used "dependencies" and not "devDependencies", because otherwise our consumers would have needed to manually install those packages. If we had just written a command line application and not expected our package to be used as a library, we might have used devDependencies.

This is only true if we are exposing declarations from those dependencies, but we aren't so we don't need to include them.

Half-Shot (changes requested):
Not convinced this is required?

jaller94:
I'm all hearing about this for the first time, so I'm trusting in your experience.

PR #222, opened 2020-09-04
Fix types: CliOpts

jaller94:
We need to separate the Opts type as a parameter from it as an internal variable.

The parameter has optional properties which the internal type assumes to be there because we have defaults for them. Using the same type makes these properties seem non-optional to the outside.


Half-Shot (approved):
Seems reasonable

PR #224, opened 2020-09-04
Types of Cli and HOWTO.md improvements

jaller94:


HOWTO.md
```diff
@@ -78,7 +78,7 @@ an application service, listen and handle incoming Matrix requests and expose a
 Open up `index.js` and add this at the bottom of the file:
 ```javascript
 var Cli = require("matrix-appservice-bridge").Cli;
```
Half-Shot, line 80: Hmm, I wonder if we should promote good practices and use const/let

Half-Shot (approved):
One comment

PR #225, opened 2020-09-08
VettedOpts for Bridge & Example: slack-starter

jaller94:
This PR corrects the types of `opts` in `new Bridge(opts)` for software that uses this library – one example project was also added.

The parameter `opts` of `Bridge` has some optional properties that are required on the internal type opts.
That's why I'm now splitting them into two types – similar to how it was done in #222.

In the long term, I would like these opts to become direct properties of `Bridge` instead of being stored in the `this.opts` object.

### DIFF
--- a/examples/slack-starter/package.json
+++ b/examples/slack-starter/package.json
@@ -1,0 +1,22 @@
+{
+  "name": "matrix-bridge-slack-starter",
+  "private": "true",
+  "version": "1.0.0",
+  "description": "",
+  "main": "lib/index.js",
+  "scripts": {
+    "start": "node ./lib/index.js",
+    "build": "tsc",
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "Matrix.org",
+  "license": "Apache-2.0",
+  "dependencies": {
+    "@types/node": "^14",
+    "@types/request": "^2.48.5",
+    "matrix-appservice-bridge": "file:../.."
+  },
+  "devDependencies": {
+    "typescript": "^4.0.2"
+  }
+}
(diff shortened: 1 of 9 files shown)


examples/slack-starter/package.json
```diff
@@ -0,0 +1,22 @@
...
+    "start": "node ./lib/index.js",
+    "build": "tsc",
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "Matrix.org",
+  "license": "Apache-2.0",
+  "dependencies": {
+    "@types/node": "^14",
+    "@types/request": "^2.48.5",
+    "matrix-appservice-bridge": "file:../.."
```
Half-Shot, line 17: Not sure this will work for github starters, we might need to reference the actual package name or is GitHub that clever?

Half-Shot: Oh nvm, I thought you were creating a github starter project here.

Half-Shot (approved):
This is really cool. I had imagined one day we might even do something like https://github.com/turt2live/matrix-bot-sdk-bot-template and create a template repo, but this is excellent as it is for now :)

examples/slack-starter/package.json
jaller94: It's meant to be run in that folder. But it could become a starter, if people copy it and reference the NPM release.

Half-Shot: Tricky, because I can easily see people copying the project to a new directory and it falling apart. However, I'm slightly more concerned with having to update the version and keep it in line with the release, so let's use this for now.

PR #226, opened 2020-09-08
Upgrade to the next version of matrix-appservice

jaller94:
Migrations:
- https://github.com/matrix-org/matrix-appservice-node/pull/27.


Half-Shot:
Happy with this. Let's merge!

PR #227, opened 2020-09-14
Add close function

### LINKED ISSUE #214: Provide method to kill bridge
Currently, it is impossible to kill the bridge via publicly accessible methods. While the appservice part can be killed via `bridge.appService.close()`, we still have the intent culling loop at https://github.com/matrix-org/matrix-appservice-bridge/blob/develop/src/bridge.js#L413 , which has to be manually killed with `clearTimeout(bridge._intentLastAccessedTimeout)`.

The bridge should instead expose a `killBridge` function that performs both of these, and whatever is needed in the future to ensure the bridge terminates.

This causes an issue with integration tests because the tests would never end with a dangling loop.

Half-Shot:
Fixes #214

### DIFF
--- a/src/bridge.ts
+++ b/src/bridge.ts
@@ -1219,9 +1219,11 @@ export class Bridge {
     public async close() {
         if (this.intentLastAccessedTimeout) {
             clearTimeout(this.intentLastAccessedTimeout);
+            this.intentLastAccessedTimeout = null;
         }
         if (this.appservice) {
             await this.appservice.close();
+            this.appservice = null;
         }
     }
 
--- a/src/bridge.ts
+++ b/src/bridge.ts
@@ -1212,6 +1212,19 @@ export class Bridge {
         return true;
     }
 
+    /**
+     * Close the appservice HTTP listener, and clear all timeouts.
+     * @returns Resolves when the appservice HTTP listener has stopped
+     */
+    public async close() {
+        if (this.intentLastAccessedTimeout) {
+            clearTimeout(this.intentLastAccessedTimeout);
+        }
+        if (this.appservice) {
+            await this.appservice.close();
+        }
+    }
+
 }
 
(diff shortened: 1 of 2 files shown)


src/bridge.ts
```diff
@@ -1212,6 +1212,19 @@ export class Bridge {
...
+     * @returns Resolves when the appservice HTTP listener has stopped
+     */
+    public async close() {
+        if (this.intentLastAccessedTimeout) {
+            clearTimeout(this.intentLastAccessedTimeout);
+        }
+        if (this.appservice) {
+            await this.appservice.close();
+        }
+    }
```
jaller94, line 1226: I think we should reset these or have a boolean flag that prevents the Bridge from being reused.
```suggestion
        if (this.intentLastAccessedTimeout) {
            clearTimeout(this.intentLastAccessedTimeout);
            this.intentLastAccessedTimeout = null;
        }
        if (this.appservice) {
            await this.appservice.close();
            this.appservice = null;
        }
    }
```

jaller94 (approved):
Generally ok, but I would wish we planned for the object being reusable or prevent it from working again.

src/bridge.ts
Half-Shot: Yeah, works for me.

jaller94:
ops. I meant undefined for appservice.

src/bridge.ts
```diff
@@ -1212,6 +1212,21 @@ export class Bridge {
...
+     * @returns Resolves when the appservice HTTP listener has stopped
+     */
+    public async close() {
+        if (this.intentLastAccessedTimeout) {
+            clearTimeout(this.intentLastAccessedTimeout);
+            this.intentLastAccessedTimeout = null;
+        }
+        if (this.appservice) {
+            await this.appservice.close();
+            this.appservice = null;
```
jaller94, line 1226:
```suggestion
            this.appservice = undefined;
```

PR #231, opened 2020-09-15
Add support for encryption via pantalaimon

### LINKED ISSUE #230: Encrypted bridges
This project involves setting up matrix-appservice-bridge to work with pantalaimon in such a way that encryption can be enabled for any bridge with no additional changes to the bridge logic.

Related PRs:
- ~Add 'usingE2EProxy' to allow users of pantialaimon to use the SDK https://github.com/matrix-org/matrix-js-sdk/pull/1469~
- ~Allow appservice users to /login - https://github.com/matrix-org/synapse/pull/8320~
- ~Do not try to login if the user authenticated via other means - https://github.com/matrix-org/pantalaimon/pull/69~
- matrix-appservice-bridge PR to setup syncing etc - https://github.com/matrix-org/matrix-appservice-bridge/pull/231
  - https://github.com/matrix-org/matrix-doc/pull/2778
- matrix-appservice-slack PRs
  - ~Types PR dependency matrix-org/matrix-appservice-slack#491~
  - Encrypted bridge bits matrix-org/matrix-appservice-slack#493

Next Phase PRs:
- Synapse to support sending EDUs down the transaction stream https://github.com/matrix-org/synapse/pull/8366
  - MSC ?

Half-Shot:
This PR enables support for encryption for all bridges, designed to be a drop in upgrade and requiring a few options to be set. 

Consequently there is a fair amount of callback fiddling and jumping around the stack to ensure that `getIntent()` continues to not make any calls until the user requires it, and that the bridge logs users in so they can interact with pan.

The meat of the changes lie in two places:
- Intent has had it's `ensureRegistered` function modified to also set up encryption for the client. This involves causing it to fetch a `session` from a datastore, or log in otherwise.
- A new class has appeared! `EncryptedEventBroker`. The broker tries to ensure that when an event arrives across the AS stream, the event is caught, and the correct event is fetched over sync and passed to onEvent. This involves some dances around retaining events temporarily in memory, and ensuring that only one event is ever handled.
- Also added support for Typing/Presence/ReadReceipts **from Matrix** by pulling the info from /sync. This won't work for non-encryption users, but `onEphemeralEvent` can be used to handle these.

This PR relies on a few dependencies, which are noted in issue #230.

This PR will likely be the last thing to be merged, so this PR fixes #230.


src/components/membership-cache.ts
```diff
@@ -72,4 +72,21 @@ export class MembershipCache {
     isUserRegistered(userId: string) {
         return this.registeredUsers.has(userId);
     }
+
+    public getMembersForRoom(roomId: string, filterFor?: UserMembership) {
+        if (!this.membershipMap[roomId]) {
+            return null;
+        }
+        // Clone so as not to alter the existing map.
```
jaller94, line 80: Comment seems outdated.

src/components/encryption.ts
```diff
@@ -0,0 +1,275 @@
...
+        const intent = this.getIntent(userId);
+        await intent.ensureRegistered();
+        const matrixClient = intent.getClient();
+        matrixClient.on("event", this.onSyncEvent.bind(this));
+        matrixClient.on("error", (err: Error) => {
+            log.error(`${userId} client error:`, err);
+        });
+        matrixClient.on("RoomMember.typing", (event: TypingEvent) => this.onTyping(userId, event));
+        matrixClient.on("Room.receipt", (event: ReadReceiptEvent) => this.onReceipt(userId, event));
+        matrixClient.on("User.presence", this.onPresence.bind(this));
```
jaller94, line 256: Some methods need .bind() but onTyping and onReceipt don't?
Those methods use `this` to in their code.

src/components/event-types.ts
```diff
@@ -0,0 +1,59 @@
...
+            }
+        }
+    }
+    type: "m.receipt";
+    // eslint-disable-next-line camelcase
+    room_id: string;
+}
+
+export interface PresenceEvent {
+    content: {
```
jaller94, line 44: If the bridge shall be able to deal with home servers that are not spec-compliant, all properties would be marked optional and tested to have the right type (including the root object.

src/components/intent.ts
```diff
@@ -41,6 +45,12 @@ export interface IntentOpts {
     dontJoin?: boolean;
     enablePresence?: boolean;
     registered?: boolean;
+    encryption?: {
+        sessionPromise: Promise<ClientEncryptionSession|null>;
+        sessionCreatedCallback: (session: ClientEncryptionSession) => Promise<void>;
+        ensureSyncingCallback: () => Promise<void>;
```
jaller94, line 51: Variable naming: Having looked at all occurrences, I'm not really sure what `ensureSyncingCallback` means.

src/components/intent.ts
```diff
@@ -827,23 +850,87 @@ export class Intent {
...
-            const res = await this.botClient.register(localpart);
-            this.opts.registered = true;
-            return res;
-        }
-        catch (err) {
-            if (err.errcode === "M_USER_IN_USE") {
+        let registerRes;
+        const userId: string = this.client.credentials.userId;
+        if (!this.opts.registered) {
+            const localpart = new MatrixUser(userId).localpart;
```
jaller94, line 877: I wish there was a static function on MatrixUser to get the localpart instead of creating an object just to throw it away. Also, the priority of new is basically never talked about. I first read this as `new (MatrixUser(userId).localpart)`
```suggestion
            const localpart = (new MatrixUser(userId)).localpart;
```

src/components/intent.ts
```diff
@@ -827,23 +850,87 @@ export class Intent {
...
+            this.readyPromise = (async () => {
+                log.debug("ensureRegistered: Attempting encrypted login");
+                // Login as the user
+                const result = await this.loginForEncryptedClient();
+                session = {
+                    userId,
+                    ...result,
+                };
+                if (this.encryption) {
+                    this.encryption.sessionPromise = Promise.resolve(session);
```
jaller94, line 925: I don't understand what's happening here.
The `encryption` object is coming from opts, so I guess the parent of this object will later check this promise and get the value???

Why is this a promise and not the value?

src/bridge.ts
```diff
@@ -998,12 +983,23 @@ export class Bridge {
...
+        const client = this.clientFactory.getClientAs(userId, request, !!this.opts.bridgeEncryption);
         const clientIntentOpts: IntentOpts = {
             backingStore: this.intentBackingStore,
             ...this.opts.intentOptions?.clients,
         };
         clientIntentOpts.registered = this.membershipCache.isUserRegistered(userId);
+        const encryptionOpts = this.opts.bridgeEncryption;
+        if (encryptionOpts) {
+            clientIntentOpts.encryption = {
+                sessionPromise: encryptionOpts.store.getStoredSession(userId),
```
jaller94, line 995: I guess here lies the answer why `sessionPromise` is a promise and not directly the session value..
because we can't await the promise here and we want to forward any errors to the parent.

jaller94 (commented):
Generally good. I should give the Slack bridge a test on Monday (or when you're done).

src/components/encryption.ts
Half-Shot: They use arrow functions which sidesteps the problem, but we should probably be consistent here.

src/components/intent.ts
Half-Shot: I've renamed it to ensureClientSyncingCallback. Essentially, it ensures the intent's client is /syncing before it performs any encrypted actions.

src/components/event-types.ts
Half-Shot: >If the bridge shall be able to deal with home servers that are not spec-compliant

We don't plan to do that at all. Writing a bridge would be a headache if that were the case :D. I've merely copied https://matrix.org/docs/spec/client_server/r0.6.0#m-presence which only has one required field.

src/components/intent.ts
Half-Shot: Have fixed the priority. Agreed I wish this was static.

src/components/intent.ts
Half-Shot: Yes, this is a tricky one. In order to not break certain APIs, we have to ensure that the *creation* of an intent is not asynchronous. This means we can't do any store lookups, any API calls or anything until the intent is used. The problem with this is it means the first opportunity to do anything is in ensureRegistered, which is always called before we perform an intent action.

So things like the session which we need to pass to the client have to be passed as a promise to unwrap once we call ensureRegistered. Ideally we wouldn't be doing this at this stage, but it's that or breaking API compatibility, which felt like it would lead to a lot of work on the other bridges. On the whole passing a promise isn't the ultimate evil.

src/bridge.ts
Half-Shot: Ah yes, see above.

Half-Shot:
Note: We need matrix-js-sdk > 8.3.0 in order for this to work, but it hasn't been released yet. In order to make the CI pass I've included the latest, but we must be careful to only release a version of the bridge once the dependency has landed.

src/components/encryption.ts
```diff
@@ -0,0 +1,283 @@
...
+     */
+    public async onASEvent(event: WeakEvent) {
+        if (event.type === "m.room.member" && event.state_key && event.content.membership) {
+            const existingSyncUser = this.userForRoom.get(event.room_id);
+            if (existingSyncUser === event.state_key && event.content.membership !== "join") {
+                // User has left the room (or are banned/invited), they are no longer our sync targets.
+                this.userForRoom.delete(event.room_id);
+                // TODO: Stop syncing entirely?
+            }
+            // TODO: If join, should we pre-emptively sync.
```
jaller94, line 120: Are you aware that you left some TODOs here?

jaller94 (approved):
I don't see anything wrong with this.

src/components/encryption.ts
Half-Shot: Having looked at this, neither seem trivial to implement and can probably wait for another time.

Half-Shot:
Shall merge once matrix-js-sdk has it's RC out.

PR #232, opened 2020-09-16
Allow info to be null on getProfileInfo

Half-Shot:
We use this in various places on the Slack bridge to get the full profile of a user.

### DIFF
--- a/src/components/intent.ts
+++ b/src/components/intent.ts
@@ -520,7 +520,7 @@ export class Intent {
      * @return A Promise that resolves with the requested user's profile
      * information
      */
-    public async getProfileInfo(userId: string, info: string, useCache=true) {
+    public async getProfileInfo(userId: string, info: "avatar_url"|"displayname"|null = null, useCache = true) {
         await this._ensureRegistered();
         if (useCache) {
             return this._requestCaches.profile.get(`${userId}:${info}`, userId, info);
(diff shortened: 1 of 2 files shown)


src/components/intent.ts
```diff
@@ -520,7 +520,7 @@ export class Intent {
      * @return A Promise that resolves with the requested user's profile
      * information
      */
-    public async getProfileInfo(userId: string, info: string, useCache=true) {
+    public async getProfileInfo(userId: string, info: "avatar_url"|"displayname"|null = null, useCache = true) {
```
jaller94, line 523: Are "avatar_url"|"displayname" the only values that `profile.get()` supports?
TypeScript becomes a pain if Types are incorrectly too strict.

But if they are the only reasonable values, great to help bridges using this library to spell them correctly.

jaller94 (approved):
Looks good to me, however, I'm concerned that the two string may be too specific.

jaller94 (changes requested):
Ah. the builds are failing.

src/components/intent.ts
Half-Shot: The spec doesn't allow arbitrary keys, and it's been that way for as long as Matrix has been around. We can update this if the spec changes at some point.

jaller94 (approved):
giphub

PR #233, opened 2020-09-24
Use actual homeserver URL for the bridge bot

Half-Shot:
The bridge bot cannot sync and should never encrypt its messages. Ensure the bridge bot goes via the homeserver URL, while ghosts use the E2E proxy.


jaller94 (approved):
I wondered if `this.opts.bridgeEncryption` has a `.enabled` boolean, but that does not seem to be the case,

so it's all good. :)

PR #236, opened 2020-09-28
Weak event should allow state_key and unsigned to be optional

### LINKED ISSUE #235: WeakEvent should not require state_key,unsigned to be defined
Both are technically optional values

Half-Shot:
Fixes #235


jaller94 (approved):
Looks good to me from a code perspective.
Gotta brush up on what those properties mean in the Matrix Spec.

PR #238, opened 2020-10-01
Add support for ephemeral events from the AS API 

Half-Shot:
Depends on https://github.com/matrix-org/matrix-appservice-node/pull/32, https://github.com/matrix-org/synapse/pull/8437

Most of the interesting things are in those two PRs, but this means we can handle ephemeral data from the AS api rather than just /sync.

### DIFF
--- a/src/bridge.ts
+++ b/src/bridge.ts
@@ -577,7 +577,8 @@ export class Bridge {
                 this.membershipCache,
                 this.appServiceBot,
                 this.onEvent.bind(this),
-                this.onEphemeralEvent.bind(this),
+                // If the bridge supports pushEphemeral, don't use sync data.
+                !this.registration.pushEphemeral ? this.onEphemeralEvent.bind(this) : undefined,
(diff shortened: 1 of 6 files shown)


src/bridge.ts
```diff
@@ -634,6 +635,9 @@ export class Bridge {
             }
             return undefined;
         });
+        this.appservice.on("ephemeral", async (event) =>
+            this.onEphemeralEvent(event as unknown as EphemeralEvent)
```
jaller94, line 639: Is this really guaranteed to be an `EphemeralEvent` or is `as unknown as X` the new `any`?

Half-Shot: It's guaranteed. `matrix-appservice-node` doesn't have types and will send us a Record<string, unknown> but that doesn't cast to a `EphemeralEvent` easily. The homeserver will only send us these events however.

jaller94 (approved):
Looks good to me.

PR #241, opened 2020-10-03
Return roomId on join

### LINKED ISSUE #239: intent.join should return the room_id
The bifrost bridge often has to join aliases, and we'd like to get the resulting room_ids from them.

Half-Shot:
Fix #239


jaller94 (approved):
Looks good to me

PR #242, opened 2020-10-05
Return event_id when sending a message

Half-Shot:
This should be non-breaking, as this only updates the types.


jaller94 (approved):
Looks good to me

PR #244, opened 2020-10-07
Add DebugAPI component

Half-Shot:
This is a generic debug API component to add bridge development and support. It's intended to replace the HTTP portions of https://github.com/matrix-org/matrix-appservice-irc/blob/develop/src/DebugApi.ts so that other bridges may set up their own debug routes.

It also allows developers to start and stop the inspector on a configured port using a curl command, rather than `kill`, which is awkward for Docker containers.


src/components/debug-api.ts
```diff
@@ -0,0 +1,57 @@
...
+        res.status(200).send({})
+    }
+
+    private onInspectorStop(_: Request, res: Response) {
+        if (!this.opts.inspector) {
+            res.status(500).send({"error": "Inspector not configured"});
+            return;
+        }
+        inspector.close();
+        res.status(200).send({})
```
jaller94, line 55: Ensuring readability and consistency (rly doing ESLint's job here)
```suggestion
        res.status(200).send({});
    }

    private onInspectorStop(_: Request, res: Response) {
        if (!this.opts.inspector) {
            res.status(500).send({"error": "Inspector not configured"});
            return;
        }
        inspector.close();
        res.status(200).send({});
```

src/components/debug-api.ts
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    public addRoute(method: "get"|"post"|"put"|"delete", path: string,
+        callback: (req: Request, res: Response) => void) {
+        this.app[method](path, callback);
+    }
+
+    private onInspectorStart(_: Request, res: Response) {
+        if (!this.opts.inspector) {
+            res.status(500).send({"error": "Inspector not configured"});
```
jaller94, line 42: An unconfigured feature shouldn't result in a HTTP 500 Server error.
How about 400 Bad request or 404 for pretending that the endpoint isn't there?

src/components/debug-api.ts
```diff
@@ -0,0 +1,57 @@
...
+            return;
+        }
+        inspector.open(this.opts.inspector.port, this.opts.inspector.host || this.opts.host);
+        res.status(200).send({})
+    }
+
+    private onInspectorStop(_: Request, res: Response) {
+        if (!this.opts.inspector) {
+            res.status(500).send({"error": "Inspector not configured"});
```
jaller94, line 51: An unconfigured feature shouldn't result in a HTTP 500 Server error.
How about 400 Bad request or 404 for pretending that the endpoint isn't there?

jaller94 (dismissed):
Looks good. I have some ideas for potential improvements though.

Half-Shot:
I don't think we're pursuing this one.

PR #247, opened 2020-10-08
Fix alias queries not arriving through the bridge

### LINKED ISSUE #245: onAliasQuery will break if onUserQuery is not defined in controller
Dodgy if statement

Half-Shot:
Fixes #245


jaller94 (approved):
Ah, the function isn't used in that method.
Looks good to me.

PR #248, opened 2020-10-09
add a function to set a user's custom displayname in a room

uhoreg:


src/components/intent.ts
```diff
@@ -575,6 +575,17 @@ export class Intent {
         return this.client.setAvatarUrl(url);
     }
 
+    public async setRoomUserDisplayName(roomId: string, name: string) {
```
Half-Shot, line 578: I'd make this generic and allow people to set their avatar too (possibly with a key: `key: "displayname"|"avatar_url", value: string`)?

src/components/intent.ts
```diff
@@ -575,6 +575,17 @@ export class Intent {
...
+        // FIXME: the content should be based on the current membership event
+        // (do we have this cached somewhere?).  Also, do nothing if the given
```
Half-Shot, line 580: `this.opts.backingStore.getMembership(roomId, userId)` will get you the membership key

Half-Shot (commented):
Some thoughts

Half-Shot (approved):
Looks great. Thank you!

PR #249, opened 2020-10-12
Fix issue where ghosts would not be registered

Half-Shot:
This function tries to determine if a user should be /registered, and only calls the method if it thinks it needs to be done. We also want to force registration sometimes (e.g. for the bridge bot), and we also want to do extra steps for encryption enabled users. This causes a bit of gnarly logic to determine if we need to skip registration.

Previously, we were seeing that users who were NOT registered were not being registered due to encryption not being enabled and forceRegister being false.

### DIFF
--- a/changelog.d/249.bugfix
+++ b/changelog.d/249.bugfix
@@ -1,0 +1,1 @@
+Fix issue where ghost users would not be registered if they've never used the bridge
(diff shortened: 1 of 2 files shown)


Half-Shot:
Having looked again. I'm unsure if this is actually broken..

PR #250, opened 2020-10-12
Don't join on kicking yourself

Half-Shot:
I've also taken the liberty to add a `userId` getter to the class, as we use it all over the place. I've also added support for folks to leave with a reason. Eventually we could use the leave call, but the JS SDK doesn't let us use a reason. You can leave with a reason in the matrix spec, so it's fine to send a kick instead for now.

### DIFF
--- a/changelog.d/250.bugfix
+++ b/changelog.d/250.bugfix
@@ -1,0 +1,1 @@
+Don't join the room when doing a self-kick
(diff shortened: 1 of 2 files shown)


jaywink (approved):
Code change looks solid.

Some reservations on kicking oneself just to show a reason, seems ... wrong :) But if the client doesn't yet support leaving with a reason then I guess it makes sense for now at least.

Half-Shot:
>Some reservations on kicking oneself just to show a reason, seems ... wrong :) But if the client doesn't yet support leaving with a reason then I guess it makes sense for now at least.

The resulting state is identical, so matrix client's wont know any difference. It's just a bit yuck from a "we're calling kick to do a leave" perspective.

PR #251, opened 2020-10-14
Membership Queue component

### LINKED ISSUE #246: Move MembershipQueue from the IRC bridge to this project

Half-Shot:
Fixes #246 

This change backports https://github.com/matrix-org/matrix-appservice-irc/blob/develop/src/util/MembershipQueue.ts to this library to be used in other bridges.

### DIFF
--- a/changelog.d/251.feature
+++ b/changelog.d/251.feature
@@ -1,0 +1,1 @@
+Add `MembershipQueue` component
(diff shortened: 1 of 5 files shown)


jaywink (approved):
Looks good to me :+1: 

The only major difference I can see is the IRC side using `QueuePool` instead of `PQueue`, but I gather there are reasons for this and the difference is taken into account in porting the code.

PR #252, opened 2020-10-15
Fix a bug where defaults would be modified on validation

Half-Shot:
This bug was causing config reloads to contain stale data


benparsons (approved):
lgtm

PR #253, opened 2020-10-20
Add pingAppserviceRoute function

Half-Shot:
This allows the appservice to check that the route from the HS->AS is accessible and working.


benparsons (approved):
lgtm

Issue #181, opened 2020-07-28
Can't send files


qwetboy10:
Unless I missed something, there doesn't seem to be a way to send images and other files with intents. Do I have to use sendEvent instead? I'd be willing to submit a pull request with the functionality, but I just want to be sure I'm not missing something before I add the functionality.

Half-Shot:
This is similar to https://github.com/matrix-org/matrix-appservice-bridge/issues/56. Currently you are expected to do `intent.getClient()` and then use `https://matrix-org.github.io/matrix-js-sdk/8.0.0/module-base-apis.MatrixBaseApis.html#uploadContent` to upload the file. 

If you want to build this functionality into the Intent class, that would be much appreciated.

PR #254, opened 2020-10-20
Add more intent calls

### LINKED ISSUE #136: Proxy more js-sdk calls through intent
I pulled this list of used calls from the `matrix-appservice-irc`. If we want to migrate over to `matrix-bot-sdk`, we will want to support this for an easy migration path.

```
.roomState()
.uploadContent()
.createAlias()
.setRoomDirectoryVisibilityAppService()
.sendStateEvent()
.getStateEvent()
```

Half-Shot:
Fixes #136, #181

### DIFF
--- a/src/components/intent.ts
+++ b/src/components/intent.ts
@@ -77,10 +77,6 @@ const returnFirstNumber = (...args: unknown[]) => {
     return 0;
 }
 
-const STATE_EVENT_TYPES = [
-    "m.room.name", "m.room.topic", "m.room.power_levels", "m.room.member",
-    "m.room.join_rules", "m.room.history_visibility"
-];
 const DEFAULT_CACHE_TTL = 90000;
 const DEFAULT_CACHE_SIZE = 1024;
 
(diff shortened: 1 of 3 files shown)


src/components/state-lookup.ts
```diff
@@ -48,7 +49,7 @@ const DEFAULT_STATE_CONCURRENCY = 4;
 
 export class StateLookup {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    private _client: any;
+    private _client: any|Intent;
```
benparsons, line 52: what does this do? is there any value to `any|*`?

src/components/intent.ts
```diff
@@ -838,9 +874,10 @@ export class Intent {
      * Ensures that the client has the required power level to post the event type.
      * @param roomId Required as power levels exist inside a room.
      * @param eventTypes The event type to check the permissions for.
+     * @param isState Are we checking for state permissions or regular event permissions.
      * @return If found, the power level event
      */
-    private async _ensureHasPowerLevelFor(roomId: string, eventType: string) {
+    private async _ensureHasPowerLevelFor(roomId: string, eventType: string, isState: boolean) {
```
benparsons, line 880: slightly instinctively I don't like bool for isState. should this be an enum? Maybe not because it will only ever be two...

benparsons (approved):
lgtm, comments not blocking

src/components/intent.ts
Half-Shot: I see what you mean, but for the foreseeable there will only be two kinds of events that power levels can control.

src/components/state-lookup.ts
Half-Shot: Not a whole lot, just helpful for future reference. Annoyingly we don't have full types for matrix-js-sdk yet, so it's called `any` all over the place.

Issue #243, opened 2020-10-06
Logging with console: "off" will cause "[winston] Attempt to write logs with no transports" when the bridge logs


Half-Shot:
If the user has explicit opted out of logs, let's actually be silent.

MadLittleMods:
For reference on a workaround that @Half-Shot shared with me, you can use the following snippet to stop the logging.

```js
const { Logging } = require('matrix-appservice-bridge');
Logging.configure({level: "silent" })
```

PR #264, opened 2020-10-30
Remove winston-daily-rotate-file side-effects if you're not using logging

MadLittleMods:
Remove `winston-daily-rotate-file` side-effects if you're not using logging

Context: https://gitlab.com/gitlab-org/gitter/webapp/-/merge_requests/2040#note_438055746

Without this change, I was seeing `TypeError: Cannot set property DailyRotateFile of #<Object> which has only a getter`
errors when using with Gitter which also has `winston@2` installed.

### DIFF
--- a/src/components/logging.ts
+++ b/src/components/logging.ts
@@ -14,10 +14,8 @@ limitations under the License.
 */
 import * as util from "util";
 import winston, { Logger, format, Logform } from "winston";
-import "winston-daily-rotate-file";
 import chalk from "chalk";
 import * as Transport from 'winston-transport';
-import { DailyRotateFile } from "winston/lib/winston/transports";
 
 type LogLevel = "debug"|"info"|"warn"|"error";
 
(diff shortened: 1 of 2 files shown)


src/components/logging.ts
```diff
@@ -121,6 +120,10 @@ class Logging {
         maxFiles: 5
     */
     configure(config: LoggerConfig = {}) {
+        // `winston-daily-rotate-file` has side-effects so we don't want to mess anyone up
+        // unless they want to use logging
+        require("winston-daily-rotate-file");
```
MadLittleMods, line 125: This did fix my problem but https://github.com/matrix-org/matrix-appservice-bridge/issues/243 is still a problem. 

Interestingly, I can still run `Logging.configure({ level: 'silent' });` which imports the problem library anyway.

I haven't tested whether `winston-daily-rotate-file` still does it's job though.

Half-Shot: I'd suggest importing it at 157, because that's when we invoke DailyRotateFile? Otherwise this is fine.

MadLittleMods: :+1: Fixed up the build as well

MadLittleMods:
Thanks for the review @Half-Shot 🦀

Half-Shot (approved):
Looks great to me :+1:

PR #267, opened 2020-10-30
Ensure encrypted events are not handled twice

### LINKED ISSUE #266: EncryptedEventBroker may emit twice for a single event when two users are syncing
Like a race somewhere.

Half-Shot:
Fixes #266 

This should address some problems with the encrypted event tracking

### DIFF
--- a/src/components/encryption.ts
+++ b/src/components/encryption.ts
@@ -131,8 +131,8 @@ export class EncryptedEventBroker {
         this.eventsPendingSync.add(event.event_id);
         const syncedEvent = this.eventsPendingAS.find((syncEvent) => syncEvent.event_id === event.event_id);
         if (syncedEvent) {
-            log.info("Got sync event before AS event");
-            this.onEvent(syncedEvent);
+            log.debug("Got sync event before AS event");
+            this.handleEvent(syncedEvent);
             return false;
         }
 
@@ -182,13 +182,11 @@ export class EncryptedEventBroker {
 
     private onSyncEvent(event: any) {
         if (!event.event.decrypted) {
-            // We only care about encrypted events, and pan appends a decrypted key to each event
-            // log.debug(`Ignoring ${event.getId()} in sync, not a encrypted event`);DedupePresence
-            // Only interested in encrypted events.
+            // We only care about encrypted events, and pan appends a decrypted key to each event.
             return;
         }
         if (!this.eventsPendingSync.has(event.getId())) {
-            log.info("Got AS event before sync event");
+            log.debug("Got AS event before sync event");
             // We weren't waiting for this event, but we might have got here too quick.
             this.eventsPendingAS.push(event.event);
             return;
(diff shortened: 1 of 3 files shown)


src/components/encryption.ts
```diff
@@ -86,16 +86,18 @@ export class EncryptedEventBroker {
         private onEphemeralEvent: ((event: EphemeralEvent) => void)|undefined,
         private getIntent: (userId: string) => Intent
         ) {
+        if (this.onEphemeralEvent) {
```
Half-Shot, line 89: If we're not handling presence, we don't need a cleanup timer.

src/components/encryption.ts
```diff
@@ -196,10 +196,23 @@ export class EncryptedEventBroker {
...
-        log.debug(`Handling ${event.getId()} through sync`);
-        this.onEvent(event.event);
+        this.handledEvents.add(`${event.room_id}:${event.event_id}`);
+
+        log.debug(`Handling ${event.event_id} through sync`);
+        this.onEvent(event);
+
+        // Delete the event from the pending list
+        this.eventsPendingSync.delete(event.event_id);
+        let index = this.eventsPendingAS.findIndex((e) => e.event_id === event.event_id);
```
Half-Shot, line 211: It's possible we might get multiple sync events stored.

src/components/encryption.ts
```diff
@@ -267,13 +280,23 @@ export class EncryptedEventBroker {
...
         if (this.onEphemeralEvent) {
             matrixClient.on("RoomMember.typing", (event: TypingEvent) => this.onTyping(userId, event));
             matrixClient.on("Room.receipt", (event: ReadReceiptEvent) => this.onReceipt(userId, event));
             matrixClient.on("User.presence", (event: PresenceEvent) => this.onPresence(event));
         }
-        const filter = new Filter(userId);
-        filter.setDefinition(SYNC_FILTER);
+        else {
+            filter.definition.presence = {
+                types: ["not.a.real.type"],
```
Half-Shot, line 292: The spec doesn't have a surefire way to turn off presence, but we can trick it by doing this.

src/components/encryption.ts
benparsons: this could be a comment in the code, not here

src/components/encryption.ts
```diff
@@ -86,16 +86,18 @@ export class EncryptedEventBroker {
         private onEphemeralEvent: ((event: EphemeralEvent) => void)|undefined,
         private getIntent: (userId: string) => Intent
         ) {
+        if (this.onEphemeralEvent) {
+            this.presenceCleanupInterval = setInterval(() => {
+                const ts = Date.now() - 30000;
```
benparsons, line 91: 30s wants to be a config var?

src/components/encryption.ts
benparsons: This can be an inline comment

src/components/encryption.ts
```diff
@@ -196,10 +196,23 @@ export class EncryptedEventBroker {
...
+        this.handledEvents.add(`${event.room_id}:${event.event_id}`);
+
+        log.debug(`Handling ${event.event_id} through sync`);
+        this.onEvent(event);
+
+        // Delete the event from the pending list
+        this.eventsPendingSync.delete(event.event_id);
+        let index = this.eventsPendingAS.findIndex((e) => e.event_id === event.event_id);
+        do {
+            this.eventsPendingAS.splice(index, 1);
```
benparsons, line 213: Would all of this be possible with `this.eventsPendingAS = this.eventsPendingAS.filter((e) => e.event_id !== event.event_id)`?

benparsons (changes requested):
questions

src/components/encryption.ts
Half-Shot: Yeah, I prefer that.

src/components/encryption.ts
Half-Shot: We should make it a const but I don't think there is any value in it, it's just a cache lifetime.

benparsons (approved):
lgtm

PR #270, opened 2020-10-31
Update matrix-appservice dependency for registration fixes

MadLittleMods:
Update `matrix-appservice` dependency for registration fixes

Includes https://github.com/matrix-org/matrix-appservice-node/pull/34

### DIFF
--- a/changelog.d/270.bugfix
+++ b/changelog.d/270.bugfix
@@ -1,0 +1,1 @@
+Update `matrix-appservice` dependency to 0.7.1 to fix a bug where `msc2409.push_ephemeral` would be required in the registration file.
--- a/package-lock.json
+++ b/package-lock.json
@@ -2954,9 +2954,9 @@
       "dev": true
     },
     "matrix-appservice": {
-      "version": "0.6.0",
-      "resolved": "https://registry.npmjs.org/matrix-appservice/-/matrix-appservice-0.6.0.tgz",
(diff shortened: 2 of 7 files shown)


MadLittleMods:
@Half-Shot Sorry TypeScript noob here. I'm not sure how to resolve some of the TypeScript stuff.

`AppServiceOutput` isn't exported from [`src/app-service-registration.ts#L10`](https://github.com/matrix-org/matrix-appservice-node/blob/69b7a92751c9f87f77660fc9e89726175238bfd9/src/app-service-registration.ts#L10) which is what I am expecting this error is referring we need to cast the `regObj` to.

https://buildkite.com/matrix-dot-org/matrix-appservice-bridge/builds/353#658cd8dd-19ef-4622-a613-e205d3776ca3
```
src/bridge.ts:552:68 - error TS2345: Argument of type 'string \| object \| undefined' is not assignable to parameter of type 'AppServiceOutput'.
--
  | Type 'undefined' is not assignable to type 'AppServiceOutput'.
  |  
  | 552             const registration = AppServiceRegistration.fromObject(regObj);
```

MadLittleMods:
Also do I need to do anything special with VSCode to see the TypeScript errors in the "Problems" section? I only see the eslint stuff currently.

Half-Shot:
I released matrix-appservice 0.7.1 to fix the issue with missing types (we needed to export the interface type), and have removed a test that expected the key to exist.

Half-Shot:
Context for @jaller94, we made the registration output format a type in matrix-appservice, so fixing up the types in the bridge lib.

PR #275, opened 2020-11-12
Add script to generate documentation

Half-Shot:


scripts/generate-docs.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+# This script will generate documentation for the current version
+
+VERSION=`python3 -c "import json; f = open('./package.json', 'r'); v = json.loads(f.read())['version']; f.close(); print(v)"`
```
jaller94, line 4: > Relying on python3 bin in a node package

```suggestion
VERSION=`/usr/bin/env node -e "console.log(require('./package.json').version)"`
```

Also, there should be some `if` and `exit` for if this fails.

scripts/generate-docs.sh
```diff
@@ -0,0 +1,20 @@
...
+git checkout gh-pages
+mv .typedoc/ $VERSION/
+sed "6i\    <li><a href=\"$VERSION/index.html\">$VERSION</a></li>" index.html > index.html.new
+rm index.html
+mv index.html.new index.html
+
+git add "$VERSION/"
+git add index.html
+git commit -m "Add documentation for $VERSION"
+git push
```
jaller94, line 20: When using automated commits and pushs some checks should have been made that the preceding commands did the expected thing (e.g. not fail or actually create the files/folders).

jaller94 (changes requested):
I'm super excited to get the documentation going, but this lacks some conditions in the bash script.
The cost of this script to go haywire isn't much but it should at least stop and exit with an error code if VERSION doesn't start with a number, yarn gendoc's return code (`$?`) isn't 0 and the branch `gh-pages` is not get checked out after the command.

scripts/generate-docs.sh
Half-Shot: Hah, fair points

scripts/generate-docs.sh
Half-Shot: Should be covered by line 6? It will fail the script if any of these return non zero.

jaller94 (approved):
Yay for auto-generating the docs.

PR #276, opened 2020-11-12
Add membership queue metrics

Half-Shot:
Fixes #259 

This PR creates 3 new metrics entries for prometheus:
- `membershipqueue_pending(type)` which will show which types of membership are waiting in the queue
- `membershipqueue_processed(type, outcome)` will show the number of completed (success & failed) membership actions
- `membershipqueue_reason(type, errcode, http_status)` will be a count of the failures for the actions
- `membershipqueue_lastage` is the age of the last processed request, in MS

Output looks good:
```
# HELP bridge_membershipqueue_pending Count of membership actions in the queue by type
# TYPE bridge_membershipqueue_pending gauge
bridge_membershipqueue_pending{type="leave"} 0
bridge_membershipqueue_pending{type="kick"} 0
bridge_membershipqueue_pending{type="join"} 0

# HELP bridge_membershipqueue_processed Count of membership actions processed by type and outcome
# TYPE bridge_membershipqueue_processed counter
bridge_membershipqueue_processed{type="leave",outcome="success"} 6
bridge_membershipqueue_processed{type="kick",outcome="success"} 4
bridge_membershipqueue_processed{type="join",outcome="success"} 1

# HELP bridge_membershipqueue_reason Count of failures to process membership, by matrix errcode and http status
# TYPE bridge_membershipqueue_reason counter

# HELP bridge_membershipqueue_lastage Gauge to measure the age of the last processed event
# TYPE bridge_membershipqueue_lastage gauge
bridge_membershipqueue_lastage 0
```

PR #277, opened 2020-11-12
Add option for a time to live for membership queue items

Half-Shot:
Needed for https://github.com/matrix-org/matrix-appservice-irc/issues/1184

Depends on #276 

This change allows us to specify a maximum ttl of a request before we just consider it stale and drop it. I've also renamed `joinDelayMs` to `actionDelayMs` because it doesn't just affect joins.

### DIFF
--- a/src/components/membership-queue.ts
+++ b/src/components/membership-queue.ts
@@ -171,9 +171,6 @@ export class MembershipQueue {
             this.pendingGauge?.dec({
                 type: item.kickUser ? "kick" : item.type
             });
-            this.pendingGauge?.dec({
-                type: item.kickUser ? "kick" : item.type
-            });
             this.processedCounter?.inc({
                 type: item.kickUser ? "kick" : item.type,
                 outcome: "success",
--- a/src/components/membership-queue.ts
+++ b/src/components/membership-queue.ts
@@ -55,6 +55,9 @@ export interface MembershipQueueOpts {
     defaultTtlMs?: number;
 }
 
+/**
+ * Default values used by the queue if not specified.
+ */
 export const DEFAULT_OPTS: MembershipQueueOptsWithDefaults = {
     concurrentRoomLimit: 8,
     maxAttempts: 10,
--- a/src/components/membership-queue.ts
+++ b/src/components/membership-queue.ts
@@ -49,10 +49,10 @@ export interface MembershipQueueOpts {
      */
     maxActionDelayMs?: number;
     /**
-     * How long a request can "live" for before it is discarded. This
-     * will override `maxAttempts`.
+     * How long a request can "live" for before it is discarded in
+     * milliseconds. This will override `maxAttempts`.
      */
-    defaultTtl?: number;
+    defaultTtlMs?: number;
 }
 
 export const DEFAULT_OPTS: MembershipQueueOptsWithDefaults = {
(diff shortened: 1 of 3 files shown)


src/components/membership-queue.ts
```diff
@@ -17,37 +18,123 @@ interface QueueUserItem {
...
+     * @deprecated Use `maxActionDelayMs`
+     */
+    maxJoinDelayMs?: number;
+    /**
+     * The maximum number of milliseconds a request may be delayed for.
+     */
+    maxActionDelayMs?: number;
+    /**
+     * How long a request can "live" for before it is discarded. This
+     * will override `maxAttempts`.
```
jaller94, line 53: This is missing a unit. With the SI base unit being seconds, I would have thought it to be seconds.
```suggestion
     * How long a request can "live" for before it is discarded in
     * milliseconds. This will override `maxAttempts`.
```

src/components/membership-queue.ts
```diff
@@ -56,15 +143,18 @@ export class MembershipQueue {
      * @param userId Leave empty to act as the bot user.
      * @param req The request entry for logging context
      * @param retry Should the request retry if it fails
+     * @param ttl How long should this request remain queued before it's discarded. Defaults to `opts.defaultTTL`
```
jaller94, line 146:
```suggestion
     * @param ttl How long should this request remain queued in milliseconds before it's discarded. Defaults to `opts.defaultTTL`
```

src/components/membership-queue.ts
```diff
@@ -76,9 +166,11 @@ export class MembershipQueue {
      * @param retry Should the request retry if it fails
      * @param reason Reason for leaving/kicking
      * @param kickUser The user to be kicked. If left blank, this will be a leave.
+     * @param ttl How long should this request remain queued before it's discarded. Defaults to `opts.defaultTTL`
```
jaller94, line 169:
```suggestion
     * @param ttl How long should this request remain queued in milliseconds before it's discarded. Defaults to `opts.defaultTTL`
```

src/components/membership-queue.ts
```diff
@@ -88,6 +180,8 @@ export class MembershipQueue {
             reason,
             kickUser,
             type: "leave",
+            ts: Date.now(),
+            ttl: ttl || this.opts.defaultTtl,
```
jaller94, line 184: If `ttl` is set to 0 this will pick the default.

```suggestion
            ttl: typeof ttl === 'number' ? ttl : this.opts.defaultTtl,
```

well... until `??` comes along.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator
Supported starting from Node 14...

jaller94 (approved):
Nice, hope we're putting an end to them jamming up the queue.
Can you add to the documentation what values for `ttl` you recommend / found to be working for our bridges?

src/components/membership-queue.ts
Half-Shot: Look forward to that in a couple of years :+1:

Half-Shot: >If ttl is set to 0 this will pick the default.

This is intended as a value of 0 would mean fail the request immediately.

Half-Shot:
> Can you add to the documentation what values for ttl you recommend / found to be working for our bridges?

I think this is covered by the default opts values?

PR #279, opened 2020-11-18
Fix uploadContent

Half-Shot:
We didn't set this flag, which meant that the raw body was being returned regardless of the other argument.

### DIFF
--- a/src/components/intent.ts
+++ b/src/components/intent.ts
@@ -707,7 +707,7 @@ export class Intent {
      */
     public async uploadContent(content: Buffer|string|ReadStream, opts: FileUploadOpts = {}): Promise<string> {
         await this.ensureRegistered();
-        return this.client.uploadContent(content, {...opts, onlyContentUri: true});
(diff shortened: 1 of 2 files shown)


Half-Shot:
hi @benparsons

benparsons (approved):
lgtm

src/components/intent.ts
```diff
@@ -707,7 +707,7 @@ export class Intent {
      */
     public async uploadContent(content: Buffer|string|ReadStream, opts: FileUploadOpts = {}): Promise<string> {
         await this.ensureRegistered();
-        return this.client.uploadContent(content, {...opts, onlyContentUri: true});
+        return this.client.uploadContent(content, {...opts, rawResponse: false, onlyContentUri: true});
```
MadLittleMods, line 710: Looks good :+1: Matches API docs https://matrix-org.github.io/matrix-js-sdk/9.1.0/module-base-apis.MatrixBaseApis.html#uploadContent

And ran into this recently in https://gitlab.com/gitterHQ/webapp/-/merge_requests/2042/diffs#note_445605432

PR #280, opened 2020-11-19
Hook into the matrix-js-sdk logger and output through winston

Half-Shot:
This fixes a longstanding bug where the js-sdk would log directly to the console which would cause spam in stdout. To solve this, we now log to winston.

### DIFF
--- a/src/components/client-factory.ts
+++ b/src/components/client-factory.ts
@@ -131,7 +131,7 @@ export class ClientFactory {
                     loggerInstance[methodName](...args);
                 }
                 else {
-                    loggerInstance.info(...args)
+                    loggerInstance.debug(...args)
                 }
             }
         }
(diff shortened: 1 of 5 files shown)


src/components/client-factory.ts
```diff
@@ -116,6 +119,34 @@ export class ClientFactory {
                 callback(err, response, body);
             });
         });
+
+        // matrix-js-sdk uses the `loglevel` logging library for it's logging
+        // but the only way to get it to log to winston is to modify the
+        // global methodFactory.
+        // To avoid potentially breaking other bridges, this is off by default.
```
benparsons, line 126: I don't understand where it's "off by default"

src/components/client-factory.ts
```diff
@@ -116,6 +119,34 @@ export class ClientFactory {
...
+        loglevel.methodFactory = (methodName, _logLevel, loggerName) => {
+            return (...args) => {
+                const loggerInstance = logging.get(`js-sdk:${loggerName}`);
+                if (methodName === "info" || methodName === "debug" ||
```
benparsons, line 130: Real nit here, but I'd remove `methodName === "info" ||` from this line, let it fall through to the else, feels more semantically correct.

benparsons (approved):
lgtm otherwise

src/components/client-factory.ts
Half-Shot: err, I think I meant to delete this comment..

benparsons (approved):
lgtm

PR #285, opened 2020-11-23
Encryption improvements

### LINKED ISSUE #284: Encrypted bridges users sometimes stop syncing
```
Nov-23 16:21:36.221 DEBUG EncryptedEventBroker Got AS event $68wefp4QhnACFnosQht4Z9jUJBD9K7k0kb3zrC8FApw
Nov-23 16:21:36.222 DEBUG Intent Checking if user @slack_astropy_foo:openastronomy.org is registered
Nov-23 16:21:36.223 INFO bridge 100.96.13.112 - - [23/Nov/2020:16:21:36 +0000] "PUT /transactions/39760?access_token=<REDACTED> HTTP/1.1" 200 2 "-" "Synapse/1.23.0"

Nov-23 16:21:36.223 DEBUG Intent ensureRegistered: Existing enc session, reusing
Nov-23 16:21:36.223 INFO EncryptedEventBroker Starting to sync @slack_astropy_foo:openastronomy.org
Nov-23 16:21:55.402 INFO bridge 100.96.3.9 - - [23/Nov/2020:16:21:55 +0000] "GET /metrics HTTP/1.1" 200 17888 "-" "Prometheus/2.16.0"

Nov-23 16:22:25.422 INFO bridge 100.96.3.9 - - [23/Nov/2020:16:22:25 +0000] "GET /metrics HTTP/1.1" 200 17886 "-" "Prometheus/2.16.0"

Nov-23 16:22:45.125 DEBUG EncryptedEventBroker Got AS event $LWPpLMeuQmqEex760cwyRO2Dw9P9j8a0BXJ6bm2zZ8o
Nov-23 16:22:45.126 DEBUG Intent Checking if user @slack_astropy_foo:openastronomy.org is registered
Nov-23 16:22:45.127 INFO bridge 100.96.13.112 - - [23/Nov/2020:16:22:45 +0000] "PUT /transactions/39761?access_token=<REDACTED> HTTP/1.1" 200 2 "-" "Synapse/1.23.0"

Nov-23 16:22:45.127 DEBUG Intent ensureRegistered: Existing enc session, reusing
Nov-23 16:22:45.127 INFO EncryptedEventBroker Starting to sync @slack_astropy_foo:openastronomy.org
Nov-23 16:22:55.401 INFO bridge 100.96.3.9 - - [23/Nov/2020:16:22:55 +0000] "GET /metrics HTTP/1.1" 200 17886 "-" "Prometheus/2.16.0"
```

Half-Shot:
Fixes #284

Depends on #295 

Sometimes we get into a state where the bridge isn't reliably syncing messages from client's and so ends up in a broken state, this aims to fix that.

### DIFF
--- a/src/bridge.ts
+++ b/src/bridge.ts
@@ -593,6 +593,7 @@ export class Bridge {
                 // If the bridge supports pushEphemeral, don't use sync data.
                 !this.registration.pushEphemeral ? this.onEphemeralEvent.bind(this) : undefined,
                 this.getIntent.bind(this),
+                this.opts.bridgeEncryption.store,
             );
         }
 
--- a/src/bridge.ts
+++ b/src/bridge.ts
@@ -705,12 +705,16 @@ export class Bridge {
             const now = Date.now();
             for (const [key, entry] of this.intents.entries()) {
                 // Do not delete intents that sync.
+                const lastAccess = now - entry.lastAccessed;
+                if (lastAccess < INTENT_CULL_EVICT_AFTER_MS) {
+                    // Intent is still in use.
+                    continue;
+                }
                 if (this.eeEventBroker?.shouldAvoidCull(entry.intent)) {
-                    return;
+                    // Intent is syncing events for encrypted rooms
+                    continue;
                 }
-                if (entry.lastAccessed + INTENT_CULL_EVICT_AFTER_MS < now) {
-                    this.intents.delete(key);
-                }
+                this.intents.delete(key);
             }
             this.intentLastAccessedTimeout = null;
             // repeat forever. We have no cancellation mechanism but we don't expect
--- a/src/components/encryption.ts
+++ b/src/components/encryption.ts
@@ -308,7 +308,7 @@ export class EncryptedEventBroker {
         const filter = new Filter(userId);
         filter.setDefinition(SYNC_FILTER);
         if (this.onEphemeralEvent) {
-            // This means that ephemeral event support ISN'T enabled for this bridge, so 
+            // This means that ephemeral event support ISN'T enabled for this bridge, so
             // we rely on sync to handle events
             matrixClient.on("RoomMember.typing", (event: TypingEvent) => this.onTyping(userId, event));
             matrixClient.on("Room.receipt", (event: ReadReceiptEvent) => this.onReceipt(userId, event));
--- a/src/components/encryption.ts
+++ b/src/components/encryption.ts
@@ -4,8 +4,6 @@ import { WeakEvent } from "./event-types";
 import { EphemeralEvent, PresenceEvent, ReadReceiptEvent, TypingEvent } from "./event-types";
 import { Intent } from "./intent";
 import Logging from "./logging";
-import matrixcs from "matrix-js-sdk";
-import { stat } from "fs";
 
 // matrix-js-sdk lacks types
 // eslint-disable-next-line @typescript-eslint/no-var-requires
--- a/src/components/encryption.ts
+++ b/src/components/encryption.ts
@@ -5,6 +5,7 @@ import { EphemeralEvent, PresenceEvent, ReadReceiptEvent, TypingEvent } from "./
 import { Intent } from "./intent";
 import Logging from "./logging";
 import matrixcs from "matrix-js-sdk";
+import { stat } from "fs";
 
 // matrix-js-sdk lacks types
 // eslint-disable-next-line @typescript-eslint/no-var-requires
@@ -19,10 +20,12 @@ export interface ClientEncryptionSession {
     userId: string;
     deviceId: string;
     accessToken: string;
+    syncToken: string|null;
 }
 export interface ClientEncryptionStore {
     getStoredSession(userId: string): Promise<ClientEncryptionSession|null>;
     setStoredSession(session: ClientEncryptionSession): Promise<void>;
+    updateSyncToken(userId: string, token: string): Promise<void>;
 }
 
 const SYNC_FILTER = {
@@ -86,7 +89,8 @@ export class EncryptedEventBroker {
         private asBot: AppServiceBot,
         private onEvent: (weakEvent: WeakEvent) => void,
         private onEphemeralEvent: ((event: EphemeralEvent) => void)|undefined,
-        private getIntent: (userId: string) => Intent
+        private getIntent: (userId: string) => Intent,
+        private store: ClientEncryptionStore,
         ) {
         if (this.onEphemeralEvent) {
             // Only cleanup presence if we're handling it.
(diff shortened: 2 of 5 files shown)


src/bridge.ts
```diff
@@ -705,12 +706,16 @@ export class Bridge {
             const now = Date.now();
             for (const [key, entry] of this.intents.entries()) {
                 // Do not delete intents that sync.
-                if (this.eeEventBroker?.shouldAvoidCull(entry.intent)) {
```
Half-Shot, line 708: In the previous version of the code, we ended up never culling any Intents because of an erronous `return` statement (I'm good at those).

This change means that Intent jobs are cleaned up if they aren't used after INTENT_CULL_EVICT_AFTER_MS and the eeEventBroker is instantiated AND needs the user kept alive (e.g. they are the transport for encrypted messages in a room)

src/components/encryption.ts
```diff
@@ -18,10 +20,12 @@ export interface ClientEncryptionSession {
     userId: string;
     deviceId: string;
     accessToken: string;
+    syncToken: string|null;
```
Half-Shot, line 21: We now store the sync token to avoid doing an initial sync on startup, which was causing unbounded amounts of memory to be used.

src/components/encryption.ts
```diff
@@ -107,7 +112,7 @@ export class EncryptedEventBroker {
     // We should probably make these LRUs eventually
     private eventsPendingAS: WeakEvent[] = [];
 
-    private syncingClients = new Set<any>();
+    private syncingClients = new Map<string, any>();
```
Half-Shot, line 113: We now store the userId => client. The client from matrix-js-sdk doesn't have a type :(

src/components/encryption.ts
```diff
@@ -142,6 +147,8 @@ export class EncryptedEventBroker {
         const existingUserForRoom = this.userForRoom.get(event.room_id);
         if (existingUserForRoom) {
             log.debug(`${existingUserForRoom} is listening for ${event.event_id}`);
+            // XXX: Sometimes the sync stops working, calling this will wake it up.
+            await this.startSyncingUser(existingUserForRoom);
```
Half-Shot, line 149: This is unfortunate but sometimes even though a user is syncing for messages, they stop working. Calling this function *checks* that the user is syncing.

src/components/encryption.ts
```diff
@@ -170,15 +177,18 @@ export class EncryptedEventBroker {
 
         const existingUser = membersForRoom.find((u) => [...this.userForRoom.values()].includes(u));
         if (existingUser) {
-            log.error(`${event.room_id} will be synced by ${existingUser}`);
+            log.debug(`${event.room_id} will be synced by ${existingUser}`);
```
Half-Shot, line 178: Less noise

src/components/encryption.ts
```diff
@@ -261,20 +271,36 @@ export class EncryptedEventBroker {
         this.onEphemeralEvent(presenceEv);
     }
 
+    private async onSync(userId: string, state: string, data: {nextSyncToken: string; catchingUp: boolean;}) {
```
Half-Shot, line 272: Update the stored sync token on insert

src/components/intent.ts
```diff
@@ -375,9 +376,18 @@ export class Intent {
         // eslint-disable-next-line camelcase
         : Promise<{event_id: string}> {
         if (this.encryption) {
-            // We *need* to sync before we can send a message.
-            await this.ensureRegistered();
-            await this.encryption.ensureClientSyncingCallback();
+            let encrypted = false;
```
Half-Shot, line 379: We previously started syncing any client that sent a message to a room, but that's inefficent if the room is unencrypted as we do not need to sync.

This code checks to see if a room is encrypted.

src/components/intent.ts
```diff
@@ -767,6 +803,9 @@ export class Intent {
         else if (event.type === "m.room.power_levels") {
             this._powerLevels[event.room_id] = event.content as unknown as PowerLevelContent;
         }
+        else if (event.type === "m.room.encryption") {
```
Half-Shot, line 812: Ensure we catch any rooms that get encryption turned on :)

src/components/prometheusmetrics.ts
```diff
@@ -380,11 +380,11 @@ export class PrometheusMetrics {
             // TODO: Ideally these metrics would be on a different port.
             // For now, leave this unauthenticated.
             checkToken: false,
-            handler: (_req: Request, res: Response) => {
+            handler: async (_req: Request, res: Response) => {
```
Half-Shot, line 383: This appears in #296 but I needed it to work for my encrypted bridge deployment #sorry

src/components/encryption.ts
```diff
@@ -261,26 +268,48 @@ export class EncryptedEventBroker {
...
+        log.debug(`${userId} sync is ${state}`);
+        if (state === "SYNCING" || state === "PREPARED" && data.nextSyncToken) {
```
jaller94, line 273: I assume that the `||` gets resolved first, and then `&&`. Could you add a parenthesis around one part, to explicitly show your intent?

src/components/intent.ts
```diff
@@ -699,6 +711,32 @@ export class Intent {
...
+            return this.encryptedRooms.get(roomId);
+        }
+        try {
+            const ev = await this.getStateEvent(roomId, "m.room.encryption");
+            const algo = ev.algorithm as string;
+            if (algo) {
+                this.encryptedRooms.set(roomId, true);
+            }
+            // Return false if missing.
+            return !!algo && algo;
```
jaller94, line 729:
```suggestion
            const algo = ev.algorithm as unknown;
            if (typeof algo === 'string' && algo) {
                this.encryptedRooms.set(roomId, true);
                return algo;
            }
            // Return false if missing, not a string or empty.
            return false;
```

src/components/intent.ts
```diff
@@ -699,6 +711,32 @@ export class Intent {
         return this.client.getStateEvent(roomId, eventType, stateKey);
     }
 
+    /**
+     * Check if a room is encrypted. If it is, return the algorithm.
+     * @param roomId The room ID to be checked
+     */
+    public async isRoomEncrypted(roomId: string) {
```
jaller94, line 718:
```suggestion
     * @param roomId The room ID to be checked
     * @returns The encryption algorithm or false
     */
    public async isRoomEncrypted(roomId: string): string|false {
```

src/components/prometheusmetrics.ts
jaller94: No need to be sorry. Great work!

jaller94 (changes requested):
Blindly trusting ev.algorithm to be a string is a bit to trusting from my perspective.
Everything else looks good.

src/components/intent.ts
Half-Shot: Very good. I don't know what happens to element if you set it to a number :D

PR #286, opened 2020-11-28
Fix runtime conversion

jld3103:
Fixes
```log
src/components/client-factory.ts:128:62 - error TS2731: Implicit conversion of a 'symbol' to a 'string' will fail at runtime. Consider wrapping this expression in 'String(...)'.

128                 const loggerInstance = logging.get(`js-sdk:${loggerName}`);
                                                                 ~~~~~~~~~~


Found 1 error.
```

### DIFF
--- a/src/components/client-factory.ts
+++ b/src/components/client-factory.ts
@@ -125,7 +125,7 @@ export class ClientFactory {
         // global methodFactory.
         loglevel.methodFactory = (methodName, _logLevel, loggerName) => {
             return (...args) => {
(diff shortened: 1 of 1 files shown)


Half-Shot:
Hmm, fair. Not sure why we think it's not a string. Can you do signoff and a changelog file as per https://github.com/matrix-org/matrix-appservice-bridge/blob/develop/CONTRIBUTING.md#making-pull-requests

Half-Shot:
Closed as fixed in another PR.

PR #288, opened 2020-12-02
Allow for returning roomId in onAliasQuery

MadLittleMods:
Allow for returning `roomId` in `onAliasQuery`.

Instead of needing to return `creationOpts`, you can now create the room yourself and return the new `roomId` you made.

---

MR for context where we want this feature: https://gitlab.com/gitterHQ/webapp/-/merge_requests/2088

As discussed in https://matrix.to/#/!ViClwbclMrgQwyPuvH:matrix.org/$VgYNMvbNXI8wjhwD5y9kI_vuQbgWgfF1-akf2VnAaNQ?via=half-shot.uk&via=matrix.org&via=tchncs.de

## Dev notes

```
npm run test -- spec/integ/bridge.spec.js
```

### DIFF
--- a/spec/integ/bridge.spec.js
+++ b/spec/integ/bridge.spec.js
@@ -187,6 +187,20 @@ describe("Bridge", function() {
             });
         });
 
+        it("should not create a room if roomId is returned from the function but should still store it",
+        async function() {
+            bridgeCtrl.onAliasQuery.and.returnValue({ roomId: "!abc123:bar" });
+            await bridge.run(101, {}, appService);
+
+            await appService.onAliasQuery("#foo:bar");
+
+            expect(clients["bot"].createRoom).not.toHaveBeenCalled();
+
+            const room = await bridge.getRoomStore().getMatrixRoom("!abc123:bar");
+            expect(room).toBeDefined();
+            expect(room.getId()).toEqual("!abc123:bar");
+        });
+
     it("should provision the room from the returned object", async() => {
             const provisionedRoom = {
                 creationOpts: {
--- a/src/bridge.ts
+++ b/src/bridge.ts
@@ -85,7 +85,7 @@ export interface BridgeController {
      * not supplied, no rooms will be provisioned on alias queries. Provisioned rooms
      * will automatically be stored in the associated `roomStore`. */
     onAliasQuery?: (alias: string, aliasLocalpart: string) =>
-        PossiblePromise<{creationOpts: Record<string, unknown>, remote?: RemoteRoom}|null|void>;
+        PossiblePromise<{roomId?: string, creationOpts?: Record<string, unknown>, remote?: RemoteRoom}|null|void>;
     /**
      * The bridge will invoke this function when a room has been created
      * via onAliasQuery.
@@ -1127,11 +1127,18 @@ export class Bridge {
             // Not provisioning room.
             throw Error("Not provisioning room for this alias");
         }
-        // eslint-disable-next-line camelcase
-        const createRoomResponse: {room_id: string} = await this.botClient.createRoom(
-            provisionedRoom.creationOpts
-        );
-        const roomId = createRoomResponse.room_id;
+
+        let roomId = provisionedRoom.roomId;
+        // If they didn't pass an existing `roomId` back,
+        // we expect some `creationOpts` to create a new room
+        if (!roomId) {
+            // eslint-disable-next-line camelcase
+            const createRoomResponse: {room_id: string} = await this.botClient.createRoom(
+                provisionedRoom.creationOpts
+            );
+            roomId = createRoomResponse.room_id;
+        }
+
         if (!this.opts.disableStores) {
             if (!this.roomStore) {
                 throw Error("roomStore is not ready yet");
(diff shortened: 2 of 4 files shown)


spec/integ/bridge.spec.js
```diff
@@ -187,6 +187,22 @@ describe("Bridge", function() {
...
+        async function(done) {
+            bridgeCtrl.onAliasQuery.and.returnValue({ roomId: "!abc123:bar" });
+            await bridge.run(101, {}, appService);
+
+            appService.onAliasQuery("#foo:bar");
+
+            expect(clients["bot"].createRoom).not.toHaveBeenCalled();
+
+            const room = await bridge.getRoomStore().getMatrixRoom("!abc123:bar");
+            expect(room).toBeDefined();
```
MadLittleMods, line 200: This bottom part to assert the `room` is just taken from the tests below but I am pretty confused about it testing anything.

If you actually debug this, `room` is `null` and the test seems to support both cases so that is why it passes. Seems to be a problem with all of the tests here.

MadLittleMods, line 200: This bottom part to assert the `room` is just taken from the tests below but I am pretty confused about it testing anything.

If you actually debug this, `room` is `null` and the test seems to support both cases so that is why it passes. Seems to be a problem with all of the tests here.

spec/integ/bridge.spec.js
```diff
@@ -187,6 +187,22 @@ describe("Bridge", function() {
             });
```
MadLittleMods, line 187: I am seeing linting and test errors for something I didn't change. Am I doing something wrong?

https://buildkite.com/matrix-dot-org/matrix-appservice-bridge/builds/411#dad87f0f-d7c1-4a2c-a9e6-8e9ea816d86d
```
src/components/client-factory.ts:128:62 - error TS2731: Implicit conversion of a 'symbol' to a 'string' will fail at runtime. Consider wrapping this expression in 'String(...)'.
--
  |  
  | 128                 const loggerInstance = logging.get(`js-sdk:${loggerName}`);
```

Half-Shot: This is https://github.com/matrix-org/matrix-appservice-bridge/pull/286

src/bridge.ts
```diff
@@ -85,7 +85,7 @@ export interface BridgeController {
      * not supplied, no rooms will be provisioned on alias queries. Provisioned rooms
      * will automatically be stored in the associated `roomStore`. */
     onAliasQuery?: (alias: string, aliasLocalpart: string) =>
-        PossiblePromise<{creationOpts: Record<string, unknown>, remote?: RemoteRoom}|null|void>;
+        PossiblePromise<{roomId?: string, creationOpts: Record<string, unknown>, remote?: RemoteRoom}|null|void>;
```
Half-Shot, line 88: You are no longer required to pass these in:

```suggestion
        PossiblePromise<{roomId?: string, creationOpts?: Record<string, unknown>, remote?: RemoteRoom}|null|void>;
```

spec/integ/bridge.spec.js
Half-Shot: Feels like you need to await `onAliasQuery` to ensure that the promises all run here? This might be a bug with the other tests.

spec/integ/bridge.spec.js
```diff
@@ -187,6 +187,22 @@ describe("Bridge", function() {
...
         });
 
+        it("should not create a room if roomId is returned from the function but should still store it",
+        async function(done) {
```
Half-Shot, line 191: I don't think you need `done` here. Some tests use it because they were written long before `async` was a thing.

spec/integ/bridge.spec.js
```diff
@@ -187,6 +187,22 @@ describe("Bridge", function() {
...
+            bridgeCtrl.onAliasQuery.and.returnValue({ roomId: "!abc123:bar" });
+            await bridge.run(101, {}, appService);
+
+            appService.onAliasQuery("#foo:bar");
+
+            expect(clients["bot"].createRoom).not.toHaveBeenCalled();
+
+            const room = await bridge.getRoomStore().getMatrixRoom("!abc123:bar");
+            expect(room).toBeDefined();
+            if (!room) { done(); return; }
```
Half-Shot, line 201: I'm concerned about this, if you await onAliasQuery you shouldn't need this?

spec/integ/bridge.spec.js
Half-Shot: With:
```js
            bridgeCtrl.onAliasQuery.and.returnValue({ roomId: "!abc123:bar" });
            await bridge.run(101, {}, appService);

            await appService.onAliasQuery("#foo:bar");

            expect(clients["bot"].createRoom).not.toHaveBeenCalled();

            const room = await bridge.getRoomStore().getMatrixRoom("!abc123:bar");
            expect(room).toBeDefined();
            expect(room.getId()).toEqual("!abc123:bar");
```

it seems to pass :)

MadLittleMods: Thanks for spotting that! Just copy pasting everything from the other tests. We should fix those up too in another PR

src/bridge.ts
MadLittleMods: Great spot! :+1:

spec/integ/bridge.spec.js
MadLittleMods: Created https://github.com/matrix-org/matrix-appservice-bridge/pull/289 to fix the other tests

PR #289, opened 2020-12-02
Fix onAliasQuery tests which would pass no matter if they failed

MadLittleMods:
Fix `onAliasQuery` tests which would pass no matter if they failed. Removing all of the broken `done` code in favor of `async`/`await`

Found while working on https://github.com/matrix-org/matrix-appservice-bridge/pull/288#discussion_r533950805

### DIFF
--- a/spec/integ/bridge.spec.js
+++ b/spec/integ/bridge.spec.js
@@ -169,22 +169,31 @@ describe("Bridge", function() {
 
     describe("onAliasQuery", function() {
         it("should invoke the user-supplied onAliasQuery function with the right args",
-        async function(done) {
+        async function() {
             await bridge.run(101, {}, appService);
-            appService.onAliasQuery("#foo:bar").catch(function() {}).finally(function() {
-                expect(bridgeCtrl.onAliasQuery).toHaveBeenCalledWith("#foo:bar", "foo");
-                done();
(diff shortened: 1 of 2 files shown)


spec/integ/bridge.spec.js
```diff
@@ -213,16 +222,16 @@ describe("Bridge", function() {
...
                 },
             });
-            bridge.run(101, {}, appService);
+            await bridge.run(101, {}, appService);
+
             await appService.onAliasQuery("#foo:bar");
+
             const room = await bridge.getRoomStore().getMatrixRoom("!abc123:bar");
             expect(room).toBeDefined();
-            if (!room) { done(); return; }
```
MadLittleMods, line 220: Removing all of this code where it would allowed the test to pass if things failed.

spec/integ/bridge.spec.js
```diff
@@ -213,16 +222,16 @@ describe("Bridge", function() {
                     room_alias_name: "foo",
                 },
             });
-            bridge.run(101, {}, appService);
+            await bridge.run(101, {}, appService);
```
MadLittleMods, line 225: Properly awaiting the bridge

Half-Shot (approved):

PR #295, opened 2021-01-14
chore: update packages

Half-Shot:


src/components/intent.ts
```diff
@@ -831,7 +831,8 @@ export class Intent {
         try {
             await this.ensureRegistered();
             if (dontJoin) {
-                deferredPromise.resolve();
+                // XXX: Should we return the passed in parameter if we didn't join, or empty?
```
jaller94, line 834: I think returning the `roomIdOrAlias` is ok as it imitates the success state after a join that's managed by this lib.
This is just for the public function `join`, right? There doesn't seem to be another use of the return value.
https://github.com/matrix-org/matrix-appservice-bridge/blob/a29960c58b62cdd348b7ff82b83d143395b64ae4/src/components/intent.ts#L546

And we're changing this from `void` to `string` because that's what the public function `join` promises, correct?

jaller94 (approved):
Nice upgrade to TypeScript, as well as other dependencies.

src/components/intent.ts
Half-Shot: Yup. Join should always return the room_id on success.

PR #296, opened 2021-01-15
Fix empty metrics due to prom-client@13

Half-Shot:


jaller94 (approved):
uff, nice catch

I confirmed on https://www.npmjs.com/package/prom-client that all examples use the function with `await`.

PR #299, opened 2021-01-25
Split out Bridge.run to two functions

### LINKED ISSUE #298: Allow Bridge methods to work without a running appservice listener
This means decoupling the appservice listener logic from initiating the bridge, so that for example you can construct intent objects on a worker process.

Half-Shot:
Fixes #298


jaller94 (approved):
Looks all good to me.

PR #300, opened 2021-01-25
Add "type" to membership queue metrics

### LINKED ISSUE #297: Membership Sync - Error: Added label "type" is not included in initial labelset: [ 'errcode', 'http_status' ]

Half-Shot:
Fixes #297

### DIFF
--- a/changelog.d/300.bugfix
+++ b/changelog.d/300.bugfix
@@ -1,1 +1,1 @@
-Fix a bug that would cause membership queue failures to not be reported as metrics
+Fix a bug that would cause membership queue request failures to not be reported under the `membershipqueue_reason` metric
--- a/changelog.d/300.bugfix
+++ b/changelog.d/300.bugfix
@@ -1,0 +1,1 @@
+Fix a bug that would cause membership queue failures to not be reported as metrics
--- a/src/components/membership-cache.ts
+++ b/src/components/membership-cache.ts
@@ -103,7 +103,7 @@ export class MembershipCache {
             return Object.keys(this.membershipMap[roomId]);
         }
         const members = [];
-        for (const [userId, [membership, profile]] of Object.entries(this.membershipMap[roomId])) {
+        for (const [userId, [membership]] of Object.entries(this.membershipMap[roomId])) {
             if (membership === filterFor) {
                 members.push(userId);
             }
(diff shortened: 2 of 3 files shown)


src/components/membership-cache.ts
```diff
@@ -103,7 +103,7 @@ export class MembershipCache {
             return Object.keys(this.membershipMap[roomId]);
         }
         const members = [];
-        for (const [userId, [membership, profile]] of Object.entries(this.membershipMap[roomId])) {
```
Half-Shot, line 106: Random linter error I spotted

jaller94: Why was this an error?
The value of `profile` isn't used.

src/components/membership-queue.ts
```diff
@@ -131,7 +131,7 @@ export class MembershipQueue {
         this.failureReasonCounter = metrics.addCounter({
             name: "membershipqueue_reason",
             help: "Count of failures to process membership, by matrix errcode and http status",
```
jaller94, line 133: This help lists all labels and should be updated (or have the list removed).

changelog.d/300.bugfix
```diff
@@ -0,0 +1 @@
+Fix a bug that would cause membership queue failures to not be reported as metrics
```
jaller94, line 1: I don't understand what this means. Why are failures reported "as metrics"? Do you mean "as part of the metrics"?
I also don't understand how adding the label "type" fixes some failures not being reported.

jaller94 (changes requested):

changelog.d/300.bugfix
jaller94: Update about 2): After having read the issue I found the code where "type" is being used as a label later.
I now understand that this is about handing over all labels on the initial sync of metrics.

src/components/membership-cache.ts
Half-Shot: Well, warning. The linter wasn't happy that profile wasn't used.

changelog.d/300.bugfix
Half-Shot: Sorta. Prom-client will get angry if you use a label not defined in the schema. This PR just ensures we define it. Have updated the changelog entry to make it a bit easier to see the bug?

src/components/membership-queue.ts
Half-Shot: Added type.

src/components/membership-cache.ts
jaller94: I also read this the commit wrong way, thinking that `profile` got added.
Nevermind. :see_no_evil:

jaller94 (approved):
Thanks for clarifying the change in the news file.

PR #302, opened 2021-01-25
Ensure bridge does not log by default

### LINKED ISSUE #243: Logging with console: "off" will cause "[winston] Attempt to write logs with no transports" when the bridge logs

Half-Shot:
Fix #243


jaller94 (approved):
Looks good.
I would have not guessed that we need this for Winston to be silent.

PR #304, opened 2021-01-25
Collectors can now be async

### LINKED ISSUE #139: registerBridgeGauges should call collectorFunc asyncronously
Some of our metrics (looking at you IRC bridge) can take some amount of ms to collect. We should optionally call this func and await the promise rather than calling synchronously.

Half-Shot:
Fixes #139

### DIFF
--- a/src/components/prometheusmetrics.ts
+++ b/src/components/prometheusmetrics.ts
@@ -18,7 +18,8 @@ import { AgeCounters } from "./agecounters";
 import JsSdk from "matrix-js-sdk";
 import { Request, Response } from "express";
 import { Bridge } from "..";
-type CollectorFunction = () => void;
+import Logger from "./logging";
+type CollectorFunction = () => Promise<void>|void;
 
 export interface BridgeGaugesCounts {
     matrixRoomConfigs?: number;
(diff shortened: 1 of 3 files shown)


src/components/prometheusmetrics.ts
```diff
@@ -173,13 +177,25 @@ export class PrometheusMetrics {
         });
     }
 
+    /**
+     * Fetch metrics from all configured collectors
+     */
+    public async refresh () {
```
jaller94, line 183: Does `refresh` need a lock (or wrapper returning an ongoing promise) so it cannot get run more than once at a time?

jaller94:
Back and forth in my head:
You seem to be pretty sure that all of these promises are fulfilled within seconds.
I'm a bit worried that a bridge calls long-running (or never-ending) functions on registration or refreshing.
Then again, that would be a (performance) bug in the bridge..
Making this async is still be an improvement, because other things can run in between.

src/components/prometheusmetrics.ts
Half-Shot: I think we're okay in this case. The caller should manage that, whether that be a bridge or the application requesting the /metrics endpoint. I don't think a race will cause any serious problems as you're just asking it to refresh metrics in collections, which should be RO functions.

Half-Shot:
> Then again, that would be a (performance) bug in the bridge..
Making this async is still be an improvement, because other things can run in between.

This was my thinking. A bridge should ensure that collector functions no not block for serious amounts of time, and if they do then the /metrics endpoint should show that clearly by taking longer rather than us trying to make the endpoint return faster but with potentially the wrong values.

jaller94 (approved):
Looks good to me. :+1:

PR #305, opened 2021-01-25
Add healthz endpoint

### LINKED ISSUE #228: Health and readyness endpoints
It would be cool to have a small component for reporting health and readyness to docker / kubernetes, that bridges can make use of.

Half-Shot:
Fixes #228 

In the end I couldn't find a way to separate health from ready -- the last thing the bridge does to ready up is listen for HTTP connections so there wasn't much point.

### DIFF
--- a/changelog.d/305.feature
+++ b/changelog.d/305.feature
@@ -1,0 +1,1 @@
+Add `/healthz` endpoint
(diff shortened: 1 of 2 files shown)


Half-Shot:
Oh derp, this got done in https://github.com/matrix-org/matrix-appservice-node/pull/38

PR #310, opened 2021-03-01
Invalidate the state cache when new room state appears

### LINKED ISSUE #308: ClientRequestCache does not expire caches when the state changes
We should also use QuickLRU rather than our own caching algo. Would be nice to make it configurable too.

Half-Shot:
Fixes #308

### DIFF
--- a/src/components/intent.ts
+++ b/src/components/intent.ts
@@ -806,13 +806,16 @@ export class Intent {
         // eslint-disable-next-line camelcase
         room_id: string
     }) {
-        if (!this._membershipStates || !this._powerLevels) {
-            return;
-        }
         if (event.state_key === undefined) {
             // We MUST operate on state events exclusively
             return;
         }
+        // Invalidate the state cache if anything changes in the state.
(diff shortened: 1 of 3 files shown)


src/components/intent.ts
```diff
@@ -806,13 +806,16 @@ export class Intent {
...
     }) {
-        if (!this._membershipStates || !this._powerLevels) {
-            return;
-        }
         if (event.state_key === undefined) {
             // We MUST operate on state events exclusively
             return;
         }
+        // Invalidate the state cache if anything changes in the state.
+        this._requestCaches.roomstate.invalidate(event.room_id);
```
benparsons, line 814: Isn't there a condition here where we call `invalidate()` then immediately return without doing anything?

Half-Shot: Yup, but this cache basically caches calls to intent.getRoomState. Basically every time we see a state event happen in a room, we want to invalidate that cache. Otherwise, we don't really care if this function doesn't do anything.

benparsons (approved):
lgtm

PR #315, opened 2021-03-04
Use m.login.application_service when registering

### LINKED ISSUE #314: We should provide a type to register

Half-Shot:
Fixes #314 

This has been a long time coming.

Spec is https://spec.matrix.org/unstable/application-service-api/#client-server-api-extensions


benparsons (approved):
lgtm

PR #316, opened 2021-03-23
Move to yarn (and update dependencies)

### LINKED ISSUE #307: Use yarn (and add yarn.lock)
The Element team [uses it](https://github.com/matrix-org/matrix-react-sdk/), and to be honest I find it to be a better tool than npm. We should switch over.

Half-Shot:
Fixes #307


jaller94 (changes requested):
Before we start bike-shedding about basically similar applications with minor advantages each, I'm just going to approve this.

Please update the ReadMe (and maybe other docs).

Half-Shot:
It's worth having the discussion about why we shouldn't do yarn if you feel strongly about it? As far as I'm aware, it's an improvement.

jaller94:
- In README.md: `npm run gendoc`
- In CONTRIBUTING.md: `npm run test && npm run lint`
- In package.json: The scripts `prepare` and `check`

All [other instances of "npm"](https://github.com/matrix-org/matrix-appservice-bridge/search?q=npm) seem ok.

jaller94 (approved):
I'm fine with changing this repo to yarn, because most of our bridge repos use yarn instead of npm.

PR #317, opened 2021-03-23
Export matrix-appservice

Half-Shot:
This was actually done a long time ago but got lost when we moved to TS. We have a recurring problem that every time we release a new `matrix-appservice` and `matrix-appservice-bridge`, every maintainer must update both projects otherwise Typescript will complain about a version mismatch. Let's avoid developers having to import both versions by just exporting it here.

### DIFF
--- a/changelog.d/317.feature
+++ b/changelog.d/317.feature
@@ -1,0 +1,1 @@
+Export `matrix-appservice` classes and interfaces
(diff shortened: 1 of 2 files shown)


jaller94 (approved):
So in a project importing `matrix-appservice-bridge` one doesn't need to import `matrix-appservice` manually at all, right?
It would just be a sub-dependency of this package.

Half-Shot:
Yeah, and typically this is more useful than asking a developer to add both, as they'll have to effectively keep them in sync each release. It's much easier to export one out of the other.

PR #318, opened 2021-03-23
Add Intent.ensureProfile

Half-Shot:
A quick helper function which bridges can call on the bot after startup to check the correct profile info is set.

### DIFF
--- a/src/components/intent.ts
+++ b/src/components/intent.ts
@@ -625,6 +625,25 @@ export class Intent {
         return this.client.setAvatarUrl(url);
     }
 
+    /**
+     * Ensure that the user has the given profile information set. If it does not,
+     * set it.
+     * @param displayname The displayname to set. Leave undefined to ignore.
+     * @param avatarUrl The avatar to set. Leave undefined to ignore.
+     */
+    public async ensureProfile(displayname?: string, avatarUrl?: string) {
+        if (!displayname && !avatarUrl) {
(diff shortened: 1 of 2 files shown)


src/components/intent.ts
```diff
@@ -625,6 +625,25 @@ export class Intent {
         return this.client.setAvatarUrl(url);
     }
 
+    /**
+     * Ensure that the user has the given profile information set. If it does not,
+     * set it.
+     * @param displayname The displayname to set. Leave undefined to ignore.
+     * @param avatarUrl The avatar to set. Leave undefined to ignore.
+     */
+    public async ensureProfile(displayname?: string, avatarUrl?: string) {
```
jaller94, line 634: Might be a bit too much of an edge case:
How to ensure the display name or avatarUrl is unset?

Falsy values will get ignored or cause the function to throw.

Half-Shot: Hmm, good question. In fairness the spec doesn't actually give you a way to unset a profile https://spec.matrix.org/unstable/client-server-api/#profiles so it's always been a bit troublesome. I'd probably say for now that the bridge should explicitly unset a profile key if they want to remove it rather than calling this function.

jaller94 (approved):
Looks good to me.
I recognise that unsetting (or setting an empty string) is an edge case that is also not defined in the spec.

PR #322, opened 2021-04-26
Update the typedoc config

Half-Shot:
This broke again, just needed some minor adjustment.


tsconfig.json
```diff
@@ -23,11 +23,9 @@
     "app.js"
   ],
   "typedocOptions": {
-    "mode": "library",
     "out": ".typedoc",
-    "entryPoint": "index",
+    "entryPoints": ["src/index.ts"],
     "excludePrivate": true,
-    "excludeNotExported": true,
     "exclude": ["**/index.ts"]
```
jaller94, line 29: This seems to exclude the entryPoint.
Does `entryPoints` trump `exclude`??

Half-Shot: :shrug: it seems to build the right thing with it excluded now.

PR #323, opened 2021-04-26
Fix typoed MSC name

Half-Shot:
Annoyed we didn't catch this one :laughing:


jaller94 (approved):
lol.. even with the change highlighted I didn't see the difference for more than 10 seconds.

I looked up the MSC and confirm that the number is now correct.

PR #326, opened 2021-05-03
Use the matrix-bot-sdk for Matrix API calls

Half-Shot:
This PR moves us away from matrix-js-sdk to the [matrix-bot-sdk](https://github.com/turt2live/matrix-bot-sdk). The advantages for which are:

- The bot-sdk is focused on bots and bridges, rather than clients.
- The bot-sdk is fully typed
- The bot-sdk includes existing support for appservices, which we could later use to further reduce our code footprint.

This PR only aims to default to using the matrix-bot-sdk Intent/Client object by default for Matrix API calls (inside Intent and otherwise). It does NOT attempt to replace the appservice logic such as handling incoming events, which might be solved via a future PR.

Great pains have been taken to be API compatible, where intent functions have kept the same signature and compatibility code has been added in places. In addition, we've kept the ability to dynamically create matrix-js-sdk instances via `getClient` as historically bridges use a mix of Intent and Client functions.

**However, upon merging this PR we will release a v3.0.0 of the library to make it an explicit upgrade choice by bridge developers.**


src/components/intent.ts
```diff
@@ -241,15 +250,38 @@ export class Intent {
...
     public getClient() {
-        return this.client;
+        if (this.legacyClient) {
+            return this.legacyClient;
+        }
+        if (!this.opts.getJsSdkClient) {
+            throw Error('Legacy client not available');
+        }
+        log.warn("Support for the matrix-js-sdk will be going away in a future release." +
+        "Please update occurances of Intent.getClient() and Intent.client");
```
jaller94, line 278: And change them to what? Not use the client directly?
This message lacks an explanation of what to do instead,

src/bridge.ts
```diff
@@ -1183,12 +1235,15 @@ export class Bridge {
...
+        if (roomId === undefined) {
+            roomId = await this.botIntent.botSdkIntent.underlyingClient.createRoom(
                 provisionedRoom.creationOpts
             );
-            roomId = createRoomResponse.room_id;
+        }
+
+        if (!roomId) {
+            // In theory this should never be called, but typescript isn't happy.
+            throw Error('Expected roomId to be defined');
```
jaller94, line 1246: You already checked that it is defined.
```suggestion
            throw Error('Expected roomId to be truthy');
```

I guess `underlyingClient.createRoom()` can return something other than a string?

jaller94 (approved):
Besides the comments I've added, this looks good to me.

src/components/intent.ts
Half-Shot: I've:

- Only made the warning fire once
- Made the warning fire via the `console`, so it is never missed by developers.
- Explained how to replace the calls.

src/bridge.ts
Half-Shot: It can't, but TS wasn't happy :/

PR #329, opened 2021-06-02
Validate edit events come from the same user

Half-Shot:
To avoid a user being able to fake an edited event


MadLittleMods (approved):
👍 I went over the changes again and everything looks good. We also have tests for the happy paths and rejection paths ✅

PR #330, opened 2021-06-02
Check m.room.create event on room upgrade

Half-Shot:
The room upgrade code previously did not check to see if the new room ID referred back to the old room ID when traversing an upgrade.


benparsons (approved):
lgtm

PR #338, opened 2021-07-15
Add debounceRequest function

### LINKED ISSUE #332: Utility to debounce room creation
I've found that we often create rooms asynchronously and build in debouncing logic all over the place to avoid the same room being created twice. I'd suggest we create a debouncing component that simply keys a room create promise to a unique key, to avoid creating multiple rooms.

Half-Shot:
Fixes #332

This basically allows callers to call a function on the intent multiple times and await the same outcome, which we often do for bridges.

### DIFF
--- a/src/components/intent.ts
+++ b/src/components/intent.ts
@@ -125,6 +125,7 @@ export class Intent {
         ensureClientSyncingCallback: () => Promise<void>;
     };
     private readyPromise?: Promise<unknown>;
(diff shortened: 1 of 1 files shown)


Half-Shot:
NB I might decide to split this out to a utility class.

tadzik (approved):
Looks good, but I'd expect it to ship with something else in the code already using it, so that it doesn't exist in a vacuum :)

Half-Shot:

PR #343, opened 2021-07-29
Remove custom rule handling code for room link validator

### LINKED ISSUE #341: Stop using a seperate rule file for RoomLinkValidator, now that we have hot-reloading support.

Half-Shot:
Fixes #341

### DIFF
--- a/src/bridge.ts
+++ b/src/bridge.ts
@@ -239,9 +239,7 @@ export interface BridgeOpts {
      */
     disableContext?: boolean;
     roomLinkValidation?: {
-        ruleFile?: string;
-        rules?: Rules;
-        triggerEndpoint?: boolean;
+        rules: Rules;
     };
     authenticateThirdpartyEndpoints?: boolean;
     roomUpgradeOpts?: RoomUpgradeHandlerOpts;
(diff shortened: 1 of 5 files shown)


src/components/room-link-validator.ts
```diff
@@ -64,8 +37,12 @@ export interface Rules {
  */
 export class RoomLinkValidator {
     private conflictCache: Map<string, number> = new Map();
-    private ruleFile?: string;
-    public readonly rules: Rules; // Public to allow unit tests to inspect it.
+    private internalRules: Rules; // Public to allow unit tests to inspect it.
+
+    public get rules(): Rules {
```
tadzik, line 42:
```suggestion
    private internalRules: Rules;

     // Public to allow unit tests to inspect it.
    public get rules(): Rules {
```

src/bridge.ts
```diff
@@ -1615,6 +1589,10 @@ export class Bridge {
         return Date.now() - sentTs;
     }
 
+    public updateRoomLinkValidatorRules(rules: Rules): void {
```
tadzik, line 1592: Is this meant to be manually called by the implementation? I wonder if we could sneak this in to the SIGHUP handler in cli.ts, but I don't quite see us having access to the Bridge itself from there...

Half-Shot: Yeah, that's the problem :(

The Cli and Bridge are logically separate and it's up to the bridge dev to hook one into the other. It's a bit naff, and I'm not sure it's obvious that you need to do this.

Half-Shot: The other problem is that since config files are a freeform schema, we wouldn't know which bit to hook into. I've given some information in the changelog, but we can do better.

tadzik: Right. I don't have any better idea, and I've been holding this up for too long :) Maybe some clever idea will come up eventually.

tadzik (approved):

PR #344, opened 2021-07-29
Do not default to port 8090 if explicitly opted out of

### LINKED ISSUE #313: Cli should check the config.yaml port before using the default port

Half-Shot:
Fixes #313

**This breaks the format of Cli.run**

~~We still default to `8090` if no default is given in Cli, but we now allow bridge developers to set their own default port. If the bridge dev explicitly sets `undefined` then **no** default will be set.~~

We now set the port to null if no port is provided.


tadzik (approved):
\o/

PR #345, opened 2021-07-29
Drop config from run

### LINKED ISSUE #262: We don't seem to use the config argument for Bridge.run
We should drop it, which does mean making some backwards incompatible changes.

Half-Shot:
Fixes #262

This is also a breaking change.


tadzik (approved):
Looks sound.

PR #349, opened 2021-09-01
Add ActivityTracker component

Half-Shot:
This has been surgically removed from https://github.com/Half-Shot/matrix-lastactive and added as a component. The bindings to the bridge class follow what we do in https://github.com/matrix-org/matrix-appservice-irc/blob/develop/src/bridge/IrcBridge.ts, with a hook to allow us to update the IRC bridge datastore.

Once this is merged, I'll deprecate matrix-lastactive.

As a side effect, our tests now support being written in TS.

### DIFF
--- a/changelog.d/594.feature
+++ b/changelog.d/594.feature
@@ -1,0 +1,1 @@
+Add tracking of last active Matrix users (previously maintained in https://github.com/Half-Shot/matrix-lastactive)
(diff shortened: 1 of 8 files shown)


Half-Shot:
I think I got a verbal tick at any rate

tadzik:
> I think I got a verbal tick at any rate

Correct :)

PR #350, opened 2021-09-01
Add active user tracking

tadzik:


src/bridge.ts
```diff
@@ -686,6 +693,16 @@ export class Bridge {
         this.setupIntentCulling();
 
         await this.loadDatabases();
+
+        this.userActivityTracker = new UserActivityTracker(
+            this.opts.userActivityTrackerConfig || UserActivityTrackerConfig.DEFAULT,
+            await this.botSdkAS.botClient.getSafeAccountData(BRIDGE_USER_ACTIVITY_STORAGE_KEY, { users: {} }),
```
Half-Shot, line 699: Like you said, I'd be a little worried about storing this much data in account data (despite what Teams does). Perhaps we could make this an interface for bridge developers to implement?

changelog.d/350.feature
```diff
@@ -0,0 +1 @@
+Add optional UserActivityTracker and BridgeBlocker
```
Half-Shot, line 1: Let's elaborate what these do, so the layman can tell if they want to upgrade or not.

spec/unit/userActivity.spec.js
```diff
@@ -0,0 +1,202 @@
+const { UserActivityTracker, UserActivityTrackerConfig  } = require("../../lib/components/userActivity")
```
Half-Shot, line 1: Had we managed to land https://github.com/matrix-org/matrix-appservice-bridge/pull/349, we could have made this typescripty. Ah well.

Half-Shot, line 1:
```suggestion
const { UserActivityTracker, UserActivityTrackerConfig  } = require("../..");
```

This is personal preference, but I like to pull components from the package root because that tests that we are properly exporting the components (and if we move them, this file doesn't need to change).

src/bridge.ts
```diff
@@ -110,6 +113,8 @@ export interface BridgeController {
             PossiblePromise<ThirdpartyUserResponse[]>;
         parseUser?(userid: string): PossiblePromise<ThirdpartyLocationResponse[]>;
     };
+
+    userActivityTracker?: UserActivityTracker;
```
Half-Shot, line 117: Interesting placement choice, we tend to leave the controller for callbacks only. Was there a reason you wanted it here rather than the config?

src/bridge.ts
```diff
@@ -155,6 +160,12 @@ export interface BridgeOpts {
      * no database will be created or used.
      */
     userStore?: UserBridgeStore|string;
+    /**
+     * The user activity store instance to use, or the path to the user .db file to load.
+     * A database will be created if this is not specified. If `disableStores` is set,
+     * no database will be created or used.
+     */
+    userActivityStore?: UserActivityStore|string;
```
Half-Shot, line 168: Very clever, I love how easy this is to opt in :)

src/bridge.ts
```diff
@@ -62,6 +64,7 @@ const INTENT_CULL_EVICT_AFTER_MS = 1000 * 60 * 15; // 15 minutes
 
 export const BRIDGE_PING_EVENT_TYPE = "org.matrix.bridge.ping";
 export const BRIDGE_PING_TIMEOUT_MS = 60000;
+export const BRIDGE_USER_ACTIVITY_STORAGE_KEY = "org.matrix.bridge.user_activity";
```
Half-Shot, line 67: Failing to spot where this is used?

src/bridge.ts
```diff
@@ -39,6 +40,7 @@ import { RoomLinkValidator, RoomLinkValidatorStatus, Rules } from "./components/
 import { RoomUpgradeHandler, RoomUpgradeHandlerOpts } from "./components/room-upgrade-handler";
 import { EventQueue } from "./components/event-queue";
 import * as logging from "./components/logging";
+import { UserActivityTracker } from "./components/userActivity";
```
Half-Shot, line 43: Camel case name is inconsistent with our other components

src/components/bridge-blocker.ts
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    constructor(private userLimit: number) {}
+
+    public checkLimits(users: number) {
+        log.debug(`Bridge now serving ${users} users`);
+
+        if (users > this.userLimit) {
+            if (!this._isBlocked) {
+                this.blockBridge().then(() => {
```
Half-Shot, line 32: Do we want to be nice here and catch failures? (I'd vote yes)

if not, we should warn implementers to catch their own failures.

src/components/bridge-blocker.ts
```diff
@@ -0,0 +1,54 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+import * as logging from "./logging";
+const log = logging.get("BridgeBlocker");
+
+export class BridgeBlocker {
```
Half-Shot, line 18: I'd like a bit more docstring around this class, as developers will need to know to override it.

src/components/intent.ts
```diff
@@ -464,10 +468,12 @@ export class Intent {
         await this._ensureJoined(roomId);
         await this._ensureHasPowerLevelFor(roomId, type, false);
         return this._joinGuard(roomId, async() => {
-            return {
+            const result = {
                 // eslint-disable-next-line camelcase
                 event_id: await this.botSdkIntent.underlyingClient.sendEvent(roomId, type, content),
-            }
+            };
+            this.opts.onEventSent?.(roomId, type, content, result.event_id);
```
Half-Shot, line 475: Real question, do we count state events?

src/components/userActivity.ts
```diff
@@ -0,0 +1,138 @@
...
+export interface UserActivityState {
+    dataSet: UserActivitySet;
+    changed: string[];
+    activeUsers: number;
+}
+type ChangesCallback = (state: UserActivityState) => void;
+
+const ONE_DAY = 24 * 60 * 60 * 1000;
+
+export class UserActivityTracker {
```
Half-Shot, line 67: Bit more doc-string needed here too

src/components/bridge-blocker.ts
Half-Shot, line 18: Actually, what's the expected usage pattern here? Do bridges use this class directly?

Half-Shot (commented):
Looking mostly good. I think for the public classes / functions we need some more docstring for the API docs. I am quite excited that this looks generic for all bridges though!

src/components/bridge-blocker.ts
tadzik: They might.

[IRC uses it as-is](https://github.com/matrix-org/matrix-appservice-irc/pull/1472/files#diff-2d366dcc8276d678ca71640bbe4bc4bc72320e4f54ac8c8f7f24642b1e2bf9f8R227)

[Slack extends it](https://github.com/matrix-org/matrix-appservice-slack/pull/612/files#diff-c9502d7239187f80ae0b64364678495cc78fd394448c552600f931ff8cc6d1f3R69) to disable its connection handlers as a blocking mechanism.

src/bridge.ts
tadzik: As far as the bridge (the MAB) is concerned, it is basically a callback (it doesn't utilize anything else, for now at least) – controller felt like a fitting place for something that *does* things. This used to have a different API where you passed UAT configuration in opts, but that violated DI and was silly in all sorts of ways.

changelog.d/350.feature
tadzik: How about something like "Add optional UserActivityTracker for tracking & reporting monthly active users, and BridgeBlocker allowing for locking down the bridge communications (intended to be used together)"?

spec/unit/userActivity.spec.js
tadzik: /me puts in on The List ;)

tadzik: Good call!

src/components/intent.ts
tadzik: Apparently we don't. Should we? Prior art[tm] did not, from what I can tell.

src/components/userActivity.ts
tadzik: Added in 6be73b58b0c0c6196c3d098bf8e2d4b651e9ae30

src/components/bridge-blocker.ts
tadzik: Added in 6be73b5

src/components/bridge-blocker.ts
tadzik: We're being nice now in 2a8861a7b640a61543e3481b25e09e4b53ca9b3c :)

src/bridge.ts
Half-Shot: Still the *case* ?

src/bridge.ts
Half-Shot: Seems reasonable. At some point we might want to revisit our bridge configuration as a whole, as it feels like this might apply to other things.

src/components/bridge-blocker.ts
Half-Shot: Another dumb question, does this spit out a integer to Prometheus yet telling us if the bridge is blocked?

Half-Shot (commented):
I think this is the home stretch now, a couple of Qs but then I think we can shippit

src/components/bridge-blocker.ts
tadzik: Oops. No. Will fix that :)

tadzik: ...actually, we can't easily do that without introducing an (optional) dependency on BridgeBlocker – currently we merely provide it for implementations but we don't force them to use it, or even assume that they might. Adding a `Bridge` opts/controller attribute just for the metric seems silly, so it may be better to just leave that up to the implementations.

changelog.d/350.feature
tadzik: Done in 8d9747001019021be8706865a6cd1a8a2c5d52f6

src/components/bridge-blocker.ts
tadzik: Now done in [IRC](https://github.com/matrix-org/matrix-appservice-irc/pull/1472/commits/c6aaf4d87b1258760f0c6913da8bcedba0105e60) and [Slack](https://github.com/matrix-org/matrix-appservice-slack/pull/612/commits/3ebd90a423efaf09565bbb76d4db4a7ad233d3c4)

Half-Shot (approved):

PR #355, opened 2021-09-20
Ensure getPrometheusMetrics() can be called in any order

Half-Shot:
This prevented us from calling this until we were listening for messages.

### DIFF
--- a/src/bridge.ts
+++ b/src/bridge.ts
@@ -613,6 +613,10 @@ export class Bridge {
             bindAddress: "127.0.0.1",
         });
 
+        if (this.metrics) {
+            this.metrics.registerMatrixSdkMetrics(this.botSdkAS);
+        }
+
         this.clientFactory = this.opts.clientFactory || new ClientFactory({
             url: this.opts.homeserverUrl,
             token: asToken,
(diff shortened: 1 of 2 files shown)


src/bridge.ts
```diff
@@ -1524,10 +1525,8 @@ export class Bridge {
 
         metrics.registerMatrixSdkMetrics(this.botSdkAS);
 
-        // TODO(paul): register some bridge-wide standard ones here
-
-        // In case we're called after .run()
-        if (this.appService && registerEndpoint) {
+        // This will throw if true and not called after run.
```
tadzik, line 1528: I wonder if we want to throw early if we know that it's gonna happen. Two layers from here we'll get a `Error('Cannot call addAppServicePath before calling .run()')`, and in this context the user is calling `getPrometheusMetrics` rather than `addAppServicePath` before calling `listen()` rather than `run()`.

Ideally I'd see all of these methods that require listen to be called first to be moved to a separate object that gets returned from listen(), so that API itself would prevent you from doing it wrong, but maybe that's an undertaking big enough for a 4.0.0 ;)

Half-Shot: Yeah, definitely a bad smell that we have these at all but I think best reserved for 4.0.0.

Half-Shot:
I think this way ensures we set the metrics up once, and it should now work in either order.

tadzik (approved):

PR #357, opened 2021-09-20
StateLookup refactors

Half-Shot:
Start using the `Intent` object to avoid dependence on the deprecated MatrixClient, and generally improve debug-ability.

### DIFF
--- a/src/components/state-lookup.ts
+++ b/src/components/state-lookup.ts
@@ -13,11 +13,11 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 import PQueue from "p-queue";
-import { Intent } from "..";
+import { Intent } from "./intent";
+import Logging from "./logging";
 
 interface StateLookupOpts {
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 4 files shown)


src/components/state-lookup.ts
```diff
@@ -116,16 +116,21 @@ export class StateLookup {
...
             return r;
         }
         catch (err) {
-            if (err.statusCode >= 400 && err.statusCode < 600) { // 4xx, 5xx
+            log.debug(`Failed to lookup state for room ${roomId}`, err);
+            const error = err as {statusCode?: number, message?: string};
+            if (error.message === "Failed to join room") {
+                throw err; // Could not join the room, don't retry.
+            }
+            if (error.statusCode && error.statusCode >= 400 && error.statusCode <= 599) { // 4xx, 5xx
```
tadzik, line 133: Unrelated to this PR, but shouldn't we retry `5xx`es?

tadzik (approved):
Looks alright

src/components/state-lookup.ts
Half-Shot: This PR plays it a bit safe, we could adapt it in the future.

PR #360, opened 2021-09-28
Support typed Matrix Errors

Half-Shot:
Depends on https://github.com/turt2live/matrix-bot-sdk/pull/158 (using a fork to make this PR work). Leaving in a draft until I get a review on that PR (as this will be a non-starter if upstream don't accept it).

This PR causes the bridge lib to emit a `MatrixError` error type which contains the useful bits of a matrix failure in a nice typed schema. All CS requests to Matrix that fail for an API reason (i.e. not a connection error) should emit this error type, so bridges should be able to look for it.


src/components/membership-queue.ts
```diff
@@ -279,11 +287,19 @@ export class MembershipQueue {
...
-            ex.statusCode === 403
-        );
+    private shouldRetry(ex: Error|MatrixError|{statusCode: number}, attempts: number): boolean {
+        if (attempts >= this.opts.maxAttempts) {
+            // Do not retry if we're out of attempts
+            return false;
+        }
+        if (ex instanceof MatrixError) {
+            // Only retry if the call wasn' forbidden
+            return ex.errcode !== "M_FORBIDDEN";
```
tadzik, line 297: I'd expect us to check for `M_LIMIT_EXCEEDED` here – any reason to retry others?

Half-Shot:
This has surely rotted.

PR #362, opened 2021-10-11
Add bridge_app_version metric

Half-Shot:
This PR adds a `bridge_app_version` metric, which is already present in the IRC bridge (essentially this PR is upstreaming that code).

### DIFF
--- a/src/utils/package-info.ts
+++ b/src/utils/package-info.ts
@@ -1,0 +1,23 @@
+import { join, resolve } from "path";
+let BridgeVersion: string;
+
+/**
+ * Get the current version of the bridge from the package.json file.
+ * @param packageJsonPath The path to the package.json of the bridge.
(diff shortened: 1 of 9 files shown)


src/utils/package-info.ts
```diff
@@ -0,0 +1,35 @@
...
+        const nodePackage = require(packageJsonPath);
+        BridgeVersion = nodePackage.version;
+    }
+    catch (err) {
+        BridgeVersion = "unknown";
+    }
+    return BridgeVersion;
+}
+
+console.log(getBridgeVersion("."));
```
tadzik, line 35: Is this intentionally left in?

tadzik (approved):

src/utils/package-info.ts
Half-Shot: No, I am a stupid.

PR #363, opened 2021-10-11
Bot command processing

Half-Shot:
This PR is largely a repurposing of https://github.com/Half-Shot/matrix-github/blob/master/src/BotCommands.ts. Notably, it's a class which you bind onto another class that offers command handlers. The idea being that it's then relatively trivial to add new commands, you just add metadata to a function and the rest works.

As an example: https://github.com/Half-Shot/matrix-github/blob/master/src/Connections/GithubRepo.ts#L208-L231


tadzik:
I'm porting Slack command handling to this to see how it fares in practice – will come back with proper review and feedback once I'm done with that – I don't want it to be a component that we ship and then don't use ;)

src/components/bot-commands.ts
```diff
@@ -0,0 +1,162 @@
...
+     * should **include** any whitspace E.g. `!irc `.
+     */
+    constructor(
+        instance: T,
+        private readonly prefix?: string) {
+        let content = "Commands:\n";
+        const botCommands: {[prefix: string]: BotCommandEntry<R>} = {};
+        const proto = Object.getPrototypeOf(instance);
+        Object.getOwnPropertyNames(proto).forEach(propetyKey => {
+            const b = Reflect.getMetadata(botCommandSymbol, instance, propetyKey) as BotCommandMetadata;
```
jaller94, line 82: What is this? Metadata or a BotCommand?
Please pick a better variable name.

src/components/bot-commands.ts
```diff
@@ -0,0 +1,162 @@
...
+            if (!b) {
+                // Not a bot command function.
+                return;
+            }
+            const optionalArgs = b.optionalArgs?.map((arg: string) => `[${arg}]`) || [];
+            const args = [...(b.requiredArgs || []), ...optionalArgs].join(" ");
+
+            content += ` - \`${this.prefix || ""}${b.name}\`${args && " "}${args} - ${b.help}\n`;
+            // We know that this is safe.
+            const fn = instance[propetyKey];
```
jaller94, line 92: typo
```suggestion
        Object.getOwnPropertyNames(proto).forEach(propertyKey => {
            const b = Reflect.getMetadata(botCommandSymbol, instance, propertyKey) as BotCommandMetadata;
            if (!b) {
                // Not a bot command function.
                return;
            }
            const optionalArgs = b.optionalArgs?.map((arg: string) => `[${arg}]`) || [];
            const args = [...(b.requiredArgs || []), ...optionalArgs].join(" ");

            content += ` - \`${this.prefix || ""}${b.name}\`${args && " "}${args} - ${b.help}\n`;
            // We know that this is safe.
            const fn = instance[propertyKey];
```

src/components/bot-commands.ts
```diff
@@ -0,0 +1,162 @@
...
+    public readonly helpMessage: TextualMessageEventContent;
+    private readonly botCommands: {[name: string]: BotCommandEntry<R>};
+
+    /**
+     * Construct a new command helper.
+     * @param prototype The prototype of the class to bind to for bot commands.
+     *                  It should contain at least one `BotCommand`.
+     * @param instance The instance of the above prototype to bind to for function calls.
+     * @param prefix A prefix to be stripped from commands (useful if using multiple handlers). The prefix
+     * should **include** any whitspace E.g. `!irc `.
```
jaller94, line 73:
```suggestion
     * should **include** any whitespace E.g. `!irc `.
```

tadzik:
Implemented this in Slack for test run [here](https://github.com/matrix-org/matrix-appservice-slack/compare/tadzik/bot-command-handler), and a few things stand out:

1) Commands are case-insensitive, which slack did for one admin room but not the other – is it configurable? Desirable?
2) There's no support for documenting command arguments, they only have their names shown
3) No support for named arguments, which Slack admin room took advantage of
4) No type checking for arguments (yargs used to provide that, perhaps we should at least provide a replacement?)
5) No support for detailed help (more verbose than the regular help)

Slack is a bit of an extreme case due to being yargs-based, and its commands could definitely be simplified – but perhaps employing a readymade command line parser is not that bad of an idea? Yargs was not fun to work with, but I've had good experiences with [`commander`](https://www.npmjs.com/package/commander) recently, and it would cover most of the above bases – perhaps we could plug that in as a backend?

Half-Shot:

PR #365, opened 2021-10-19
Provisioning and widgets API

Half-Shot:
This PR lands the very basics of the widgets / provisioning API, intended to be adopted by our bridge ecosystem. In this model, widgets and provisioning apis are under the same roof where provisioners get a powerful token to authenticate requests as any user, and widgets can get a token for a user by supplying an openId token. 

There are some open questions:
 - The well known implementation needs to be better, ideally it should land in the bot sdk.
 - Are there any unforeseen dangers with mixing the two systems
 - This adds more dependencies to matrix-appservice-bridge, which some people might describe as massively excessive for their needs. Can we make this somehow lighter?

I'm hoping to land this as a self contained unit, and we can then start experimenting with widgets more generally.


src/provisioning/api.ts
```diff
@@ -0,0 +1,243 @@
...
+        // TODO: Need a MUCH better impl:
+        try {
+            const wellKnown = await axios.get<MatrixServerWellKnown>(`https://${server}/.well-known/matrix/server`, {
+                validateStatus: null
+            });
+            if (wellKnown.status === 200) {
+                url = `https://${wellKnown.data["m.server"]}`;
+            }
+            else {
+                url = `https://${server}:8448`;
```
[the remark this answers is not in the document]
Half-Shot: Matrix servers default to 8448 as per https://spec.matrix.org/v1.1/server-server-api/#resolving-server-names. Though fair warning that this section needs improving, so I'm going to write it again.

Half-Shot:
Right, so I think this is generally ready for another review. I can only apologize massively for how much this has ballooned, but I think it's going to be a good going forwards. The highlights are:

- The API now has a sensible way to do errors, and definitely felt more ergonomic when developing the Slack bridge components.
- I've created a helper utility `MatrixHostResolver` which should provide a **complete** host resolver for Matrix Servers so that the OpenID components work better
- Generally tidied up the shop as I've gone along with my demos.

I don't have a Slack PR up yet, but I can at least provide my diff thus far if you want  to see this used in action over at https://github.com/matrix-org/matrix-appservice-slack/compare/develop...hs/invite-widget-support

spec/log.js
```diff
@@ -1,2 +0,0 @@
-module.exports.beforeEach = function(testCase) {
```
Half-Shot, line 1: I couldn't see any point to this existing, so I yeeted it.

src/utils/matrix-host-resolver.ts
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    get currentTime(): number {
+        return this.opts.currentTimeMs || Date.now();
+    }
+
+
+    private static sortSrvRecords(a: SrvRecord, b: SrvRecord): number {
+        // This algorithm is intentionally simple, as we're unlikely
+        // to encounter many Matrix servers that acatually load balance this way.
```
jaller94:
```suggestion
        // to encounter many Matrix servers that actually load balance this way.
```

src/provisioning/api.ts
```diff
@@ -0,0 +1,322 @@
...
+        if (!authHeader) {
+            throw new ApiError('No Authorization header', ErrCode.BadToken);
+        }
+        const token = authHeader.startsWith("bearer ") && authHeader.substr("bearer ".length);
+        if (!token) {
+            return;
+        }
+        const requestProv = (req as ExpRequestProvisioner);
+        if (!this.opts.provisioningToken && req.body.userId) {
+            throw new ApiError('Provisioing feature disabled', ErrCode.DisabledFeature);
```
jaller94:
```suggestion
            throw new ApiError('Provisioning feature disabled', ErrCode.DisabledFeature);
```

jaller94 (commented):
So far I haven't found any evidence for issues, but these were the ones I was looking for:
* Brute force attack vector for tokens of the Provisioning API. Could be mitigated via a firewall, fail2ban based on log lines and rate limiting.
* Probing/Manipulating internal network hosts by setting `m.sever` to an internal IP addresses. But the wellKnown responses are only available to people via the Provisioning API, right?

Half-Shot:
>Brute force attack vector for tokens of the Provisioning API. Could be mitigated via a firewall, fail2ban based on log lines and rate limiting.

Ratelimiting seems like a sensible one, beyond that I think it's down to server admins to add appropriate protection.

>Probing/Manipulating internal network hosts by setting m.sever to an internal IP addresses. But the wellKnown responses are only available to people via the Provisioning API, right?

It could be manipulated by anyone, because it's used to validate widget user's identitys via OpenID. That said, I think manipulating it would only cause the bridge to hit /_matrix/federation/v1 on internal services. I think it's better safe than sorry, so we should add the ability to exclude internal addresses.

Half-Shot:
I've added ratelimiting (off by default though, would rather enable this at the bridge author / admin's discretion). I've also added an IP ban list to prevent internal IP reaching. The defaults match Synapse's current set, though that can be overridden.

jaller94 (approved):
It's a pretty big PR and hard to grasp and check every change, but I think it's an awesome start towards the new widget system which will make provisioning so much easier. :+1: :+1:

PR #366, opened 2021-10-21
Add greedy regex for the bridge namespace

Half-Shot:
We don't use the namespace login in the bot-sdk for anything, so this should be safe to do. We don't use this namespace to inform Synapse, so there isn't a risk to security either.

A solution for https://github.com/turt2live/matrix-bot-sdk/issues/159

Hopefully helps solve https://github.com/matrix-org/matrix-appservice-irc/issues/1490

### DIFF
--- a/changelog.d/366.bugfix
+++ b/changelog.d/366.bugfix
@@ -1,0 +1,1 @@
+Fix a bug where bridges could not register multiple user namespaces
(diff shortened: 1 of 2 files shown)


tadzik (approved):
Seems to fix the issue when tested locally. Looks good to me too :)

PR #369, opened 2021-11-03
Refactor encryption code to use bot-sdk 

Half-Shot:
This code refactors our encryption broker component and intents to generally handle encrypted events better. The changes are:

- Use the bot-sdk for encryption, rather than the js-sdk
- Drop support for ephemeral events via /sync, preferring the AS route.
- Refactors the session creation code within Intent, hopefully reducing the chance of a race.
- Improves the example encryption app.


src/components/intent.ts
```diff
@@ -1180,51 +1185,81 @@ export class Intent {
...
-                    this.encryption.sessionPromise = Promise.resolve(session);
-                }
-                await this.encryption?.sessionCreatedCallback(session);
-            })();
-            await this.readyPromise;
+
+        if (!session) {
+            // No session in the store, attempt a login.
+            log.debug("ensureRegistered: Attempting encrypted login");
+            // Login as the user
```
jaller94, line 1242: Login is the noun.
```suggestion
            // Log in as the user
```

src/components/intent.ts
```diff
@@ -129,8 +118,13 @@ export class Intent {
         sessionPromise: Promise<ClientEncryptionSession|null>;
         sessionCreatedCallback: (session: ClientEncryptionSession) => Promise<void>;
         ensureClientSyncingCallback: () => Promise<void>;
+        origianlHomeserver: string
     };
-    private readyPromise?: Promise<unknown>;
+    private encryptionReadyPromise?: Promise<void>;
+
+    // A client that talks directly to the homeserver, bypassing pan.
```
jaller94, line 125: To reduce the knowledge one needs to have to understand this comment, I'd recommend not abbreviating pantalaimon.
```suggestion
    // A client that talks directly to the homeserver, bypassing pantalaimon.
```

src/components/intent.ts
```diff
@@ -1180,51 +1185,81 @@ export class Intent {
             }
         }
 
+        // NOTES:
+        // Still seeing errors about invalid access tokens.
+        // Still seeing the server attempt to send stuff to pan without sync being enabled.
```
jaller94, line 1190:
```suggestion
        // Still seeing the server attempt to send stuff to pantalaimon without sync being enabled.
```

src/components/encryption.ts
```diff
@@ -189,163 +185,139 @@ export class EncryptedEventBroker {
         return false;
     }
 
-    private onSyncEvent(event: any) {
-        if (!event.event.decrypted) {
+    private onSyncEvent(roomId: string, event: PanWeakEvent): void {
+        log.info("BLARGH", roomId, event);
+        if (!event.decrypted) {
             // We only care about encrypted events, and pan appends a decrypted key to each event.
```
jaller94, line 191:
```suggestion
            // We only care about encrypted events, and pantalaimon appends a decrypted key to each event.
```

src/components/encryption.ts
```diff
@@ -189,163 +185,139 @@ export class EncryptedEventBroker {
         return false;
     }
 
-    private onSyncEvent(event: any) {
-        if (!event.event.decrypted) {
+    private onSyncEvent(roomId: string, event: PanWeakEvent): void {
+        log.info("BLARGH", roomId, event);
```
jaller94, line 189:
```suggestion
```
or
```suggestion
        log.debug("onSyncEvent", roomId, event);
```

src/components/encryption.ts
```diff
@@ -189,163 +185,139 @@ export class EncryptedEventBroker {
...
+        if (!event.decrypted) {
             // We only care about encrypted events, and pan appends a decrypted key to each event.
             return;
         }
-        if (!this.eventsPendingSync.has(event.getId())) {
+        // TODO: Do we need this?
```
jaller94, line 194: Just pointing this out: You've left a TODO here.

src/components/intent.ts
```diff
@@ -1180,51 +1185,81 @@ export class Intent {
...
+            // Failed to ready up - fall through and try again.
+        }
+    }
 
-        // Encryption enabled, check if we have a session.
+    private async getEncryptedSession(): Promise<void> {
+        if (!this.encryption) {
+            throw Error('Cannot call getEncryptedSession without enabling encryption');
+        }
+        // We've not got a session so let's see if the store has one.
```
jaller94, line 1221:
```suggestion
        // We've got no session so let's see if the store has one.
```

src/components/intent.ts
```diff
@@ -1180,51 +1185,81 @@ export class Intent {
...
-            log.debug("ensureRegistered: ready promise ongoing");
-            try {
-                // Should fall through and find the session.
-                await this.readyPromise;
-            }
-            catch (ex) {
-                log.debug("ensureRegistered: failed to ready", ex);
-                // Failed to ready up - fall through and try again.
-            }
+        // Past this point, we're an encryption enabled Intent.
```
jaller94, line 1198:
```suggestion
        // Past this point, we're an encryption-enabled Intent.
```

src/bridge.ts
```diff
@@ -1165,13 +1164,14 @@ export class Bridge {
...
             ...this.opts.intentOptions?.clients,
-            onEventSent: () => this.opts.controller.userActivityTracker?.updateUserActivity(userId!),
+            onEventSent: () => userId && this.opts.controller.userActivityTracker?.updateUserActivity(userId),
         };
         clientIntentOpts.registered = this.membershipCache.isUserRegistered(userId);
         const encryptionOpts = this.opts.bridgeEncryption;
         if (encryptionOpts) {
             clientIntentOpts.encryption = {
                 sessionPromise: encryptionOpts.store.getStoredSession(userId),
+                origianlHomeserver: this.opts.homeserverUrl,
```
jaller94, line 1174: typo that hopefully still can be fixed in the entire repo.
```suggestion
                originalHomeserver: this.opts.homeserverUrl,
```

jaller94 (changes requested):

Half-Shot:
I had an epiphany and decided to refactor the encryption code into a separate subclass so it's easier to understand (and less if/else-ing)

src/components/membership-cache.ts
```diff
@@ -91,11 +91,11 @@ export class MembershipCache {
      * Is a user considered registered with the homeserver.
      * @param userId A Matrix userId
      */
-    isUserRegistered(userId: string) {
+     public isUserRegistered(userId: string): boolean {
```
tadzik, line 94:
```suggestion
    public isUserRegistered(userId: string): boolean {
```
(I'm sorry)

src/components/encrypted-intent.ts
```diff
@@ -0,0 +1,237 @@
...
+     */
+    public async uploadContent(content: Buffer|string|ReadStream, opts: FileUploadOpts = {}): Promise<string> {
+        await this.ensureRegistered();
+        // Media is encrypted, since we don't know the destination room assume this media will be encrypted.
+        await this.encryptionOpts.ensureClientSyncingCallback();
+        return super.uploadContent(content, opts);
+    }
+
+    public onEvent(event: {
+        type: string,
```
tadzik, line 56: I'm suprised that RoomEvent from bot-sdk doesn't cover this (no room_id in it seemingly). I wonder if we ought to just define our type for this, since it comes up quite a bit ime.

src/components/encrypted-intent.ts
```diff
@@ -0,0 +1,237 @@
...
+                }
+            },
+        );
+        return {
+            accessToken: res.access_token as string,
+            deviceId: res.device_id as string,
+        };
+    }
+
+    public async ensureRegistered(forceRegister = false): Promise<"registered=true"|undefined> {
```
tadzik, line 91: What's the reason for it returning this awkward string rather than a true (or just void, since it rejects on failure anyway?)

src/components/encryption.ts
```diff
@@ -1,18 +1,19 @@
...
+import { MatrixClient } from "matrix-bot-sdk";
+import LRU from "@alloc/quick-lru"
 
 const log = Logging.get("EncryptedEventBroker");
 
 export const APPSERVICE_LOGIN_TYPE = "uk.half-shot.msc2778.login.application_service";
-const PRESENCE_CACHE_FOR_MS = 30000;
+const EVENT_CACHE_FOR_MS = 5 * 60000; // 5 minutes
+
+interface PanWeakEvent extends WeakEvent {
```
tadzik, line 14: I'd go for PantalaimonWeakEvent. We've got autocompletion, there's no excuse ;)

src/components/encryption.ts
```diff
@@ -357,18 +327,21 @@ export class EncryptedEventBroker {
...
+            }
         }
     }
 
-    public static supportsLoginFlow(loginFlows: {flows: {type: string}[]}) {
+    public static supportsLoginFlow(loginFlows: {flows: {type: string}[]}): boolean {
         return loginFlows.flows.find(
             flow => flow.type === APPSERVICE_LOGIN_TYPE
-        );
+        ) !== undefined;
```
tadzik, line 345: I'd go for a `!!` prefix but this is fine.

src/components/encrypted-intent.ts
```diff
@@ -0,0 +1,237 @@
...
+    ensureClientSyncingCallback: () => Promise<void>;
+    originalHomeserverUrl: string;
+}
+
+/**
+ * Implements some special handling on top of Intent to handle encrypted rooms.
+ */
+export class EncryptedIntent extends Intent {
+
+    private readonly encryptedRooms = new Map<string, string|false>();
```
tadzik, line 22: Do we care what algo it is? It is in the public API technically, but named as if it was a bool. Not sure how much of a breaking change that'd be, but I'd break it to rename it, if it's needed at all :)

src/components/encryption.ts
jaller94: I'd go for `.findIndex() !== -1` :laughing: 
So many options.

src/components/encrypted-intent.ts
Half-Shot: It's a thing that the old old sdk did and we've just carried it forward. I'm not sure if anyone actually expects it..

src/components/encrypted-intent.ts
Half-Shot: Fair, we can save some bytes here.

src/components/encryption.ts
Half-Shot: I've left it alone because I want to be contrarian :D

tadzik (approved):

PR #371, opened 2021-11-03
Do not retry membership on a 404 error

### LINKED ISSUE #370: MembershipQueue should handle MatrixHttpClient (REQ-2961) { errcode: 'M_UNKNOWN', error: 'Not a known room' }

Half-Shot:
Fixes #370  

This usually means the room doesn't exist for ex: https://github.com/matrix-org/synapse/blob/c01bc5f43d1c7d0a25f397b542ced57894395519/synapse/handlers/room_member.py#L812.

### DIFF
--- a/changelog.d/371.bugfix
+++ b/changelog.d/371.bugfix
@@ -1,0 +1,1 @@
+Do not retry membership queue operations after a 404 response.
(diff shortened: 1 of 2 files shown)


jaller94 (approved):
Looks good.

I wonder if other HTTP errror codes like 400 BAD REQUEST should also return `false`. Other than 429 Too Many Requests none of the 4xx-codes seem recoverable.

Half-Shot:
It's tricky, in the wild I've seen recoveries from generic M_UNKNOWN / 500 requests but you're right. Maybe this needs to check for explicit recoverable errors.

jaller94:
Yeah, an HTTP 500 should be recoverable, but the 4xx-errors are almost all not recoverable.

PR #373, opened 2021-11-10
Typo: GagueOpts -> GaugeOpts

jaller94:
Seems like this interface is not exported and safe to rename without breaking things.


Half-Shot (approved):
oops

PR #377, opened 2021-11-23
Update typedoc so that docs can be built

Half-Shot:
This really needs to be a CI step so we know if it breaks


Half-Shot:
I think this is relatively uncontentious.

PR #388, opened 2022-01-27
Slight refactors to provisioner

Half-Shot:
- Allow other error types to be thrown, using the interface IApiError
- Don't require the store to throw an ApiError, instead allow it to return null.
- Allow userId to be null, as some provisioning paths don't requite a userId (e.g. checking limits on the bridge).


jaller94 (approved):
Looks good to me.

PR #389, opened 2022-02-05
Use m.login.application_service

Half-Shot:
This is now part of the Matrix spec, we don't need to use the unstable prefix. Suggest the next version of this library is breaking to support this.


jaller94 (approved):
Looks good to me.
Yeah, lets definitely wait to a release that's ok to include breaking changes.

PR #393, opened 2022-02-16
Support extensible events

### LINKED ISSUE #390: Emit extensible events via matrix-events-sdk
https://github.com/matrix-org/matrix-events-sdk#readme

Half-Shot:
Closes #390 

Matrix is moving to a new model for events where by messages no longer contain one representation, but a series of fallbacks (known as extensible events) so that applications can choose the best representation.

The bridge sdk will make use of the `matrix-events-sdk` to automatically parse and represent these events, and then use a new controller callback to emit these events. This is currently unstable, and so bridges need to explicitly opt in to this behavior to use it.

See https://github.com/matrix-org/matrix-events-sdk#readme
See https://github.com/matrix-org/matrix-doc/pull/1767

### DIFF
--- a/src/components/request-factory.ts
+++ b/src/components/request-factory.ts
@@ -33,8 +33,8 @@ export class RequestFactory {
      * @param opts The options to pass to the Request constructor, if any.
      * @return A new request object
      */
-    public newRequest<T>(opts?: RequestOpts<T>) {
-        const req = new Request(opts || {data: null});
+    public newRequest<T, R extends unknown>(opts?: RequestOpts<T>): Request<T, R> {
+        const req = new Request(opts);
         req.getPromise().then((res) => {
             this._resolves.forEach((resolveFn) => {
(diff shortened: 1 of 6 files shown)


src/components/request-factory.ts
```diff
@@ -33,8 +33,8 @@ export class RequestFactory {
      * @param opts The options to pass to the Request constructor, if any.
      * @return A new request object
      */
-    public newRequest<T>(opts?: RequestOpts<T>) {
-        const req = new Request(opts || {data: null});
+    public newRequest<T, R extends unknown>(opts?: RequestOpts<T>): Request<T, R> {
```
tadzik, line 36: Nitpick: does `R extends unknown` do anything more than a regular `R`?

tadzik:
Looks good: do we have any bridge currently making use of this, so we can see how the API works out in practice?

Half-Shot:

PR #397, opened 2022-04-05
Provisioner small fixes

Half-Shot:
A few things I found annoying when working on the hookshot widget stuff:
 - `addRoute` had only a few supported methods, this has now been increased to all supported methods by `Router`
 - The openID handling code wasn't correctly concatenating the homeserver base URL and the request path.
 - The request logging felt too noisy at info, so I reduced it down to `debug`. I'm still 50/50 on that change.
 - The matrix-host-resolver would die if the well known Content-Type was not `application/json`. While we could be strict, it appears Synapse is not and we should attempt to parse strings as JSON.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,27 @@
- 3.2.0 (2021-11-23)
+4.0.0 (2022-03-31)
+==================
+
+Features
+--------
+
+- Add new `Provisioning and Widgets API`, so that bridges can provide richer integration APIs ([\#365](https://github.com/matrix-org/matrix-appservice-bridge/issues/365), [\#388](https://github.com/matrix-org/matrix-appservice-bridge/issues/388))
+
+
+Deprecations and Removals
+-------------------------
+
+- Drop support for Node 12. Support Node 14+. ([\#382](https://github.com/matrix-org/matrix-appservice-bridge/issues/382))
+
+
+Internal Changes
+----------------
+
+- Update typedoc to 0.22.9 ([\#377](https://github.com/matrix-org/matrix-appservice-bridge/issues/377))
+- Logging in as an appservice user (for encryption support) now uses `m.login.application_service` rather than the unstable prefix. This may break on homeservers that are not up to date with Matrix v1.2. ([\#389](https://github.com/matrix-org/matrix-appservice-bridge/issues/389))
+- Replace Buildkite with GitHub Actions for CI. ([\#392](https://github.com/matrix-org/matrix-appservice-bridge/issues/392))
+
+
+3.2.0 (2021-11-23)
 ===================
 
 Features
--- a/src/provisioning/api.ts
+++ b/src/provisioning/api.ts
@@ -349,7 +349,8 @@ export class ProvisioningApi {
 
         // Now do the token exchange
         try {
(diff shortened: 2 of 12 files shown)


src/provisioning/request.ts
```diff
@@ -1,16 +1,51 @@
...
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    ReqBody = any,
+    ReqQuery = ParsedQs> implements ThinRequest {
     public readonly log: LogWrapper;
     public readonly id: string;
 
-
     constructor(
-        private expressReq: {body: Body, params: Params, path?: string},
+        public readonly expressReq: Request<Params, ResBody, ReqBody, ReqQuery>,
```
Half-Shot, line 48: Exposing this publicly because I think it's probably going to end up useful eventually.

src/provisioning/api.ts
```diff
@@ -349,7 +350,8 @@ export class ProvisioningApi {
 
         // Now do the token exchange
         try {
-            const response = await axios.get<{sub: string}>(`${url}/_matrix/federation/v1/openid/userinfo`, {
+            const requestUrl = new URL("/_matrix/federation/v1/openid/userinfo", url);
```
Half-Shot, line 353:
```
> new URL("/_matrix/federation/v1/openid/userinfo", new URL("https://matrix.org"));
URL {
  href: 'https://matrix.org/_matrix/federation/v1/openid/userinfo',
  origin: 'https://matrix.org',
  protocol: 'https:',
  username: '',
  password: '',
  host: 'matrix.org',
  hostname: 'matrix.org',
  port: '',
  pathname: '/_matrix/federation/v1/openid/userinfo',
  search: '',
  searchParams: URLSearchParams {},
  hash: ''
}
```

CHANGELOG.md
```diff
@@ -1,4 +1,27 @@
- 3.2.0 (2021-11-23)
+4.0.0 (2022-03-31)
```
Half-Shot, line 1: I managed to forget to push https://github.com/matrix-org/matrix-appservice-bridge/commit/30f6e3a4f4c8e3ec0c4d08055e39845a71d34763 until after I submitted this PR, which is why the diff looks odd. This is actually on `develop` and GitHub is just confused.

src/provisioning/api.ts
tadzik: I hate URL's API, thanks for documenting it here :)

tadzik (approved):
Looks good to me.

I'm curious about this bit though:

> The matrix-host-resolver would die if the well known Content-Type was not application/json. While we could be strict, it appears Synapse is not

Is there a bug for Synapse not behaving nicely here?

Half-Shot:
>Is there a bug for Synapse not behaving nicely here?

I think Synapse is probably doing the correct thing, and in reality well known parsers should try to convert the response into JSON regardless of Content-Type. The problem here is that axios and other HTTP client libraries will not do this unprompted. The spec doesn't mandate a content type, having looked at it. https://spec.matrix.org/v1.2/server-server-api/#resolving-server-names

Half-Shot:
I did open an issue on https://github.com/matrix-org/matrix-spec/issues/1025

PR #399, opened 2022-05-03
Fix mention of MSC2666 in error message

Half-Shot:
This should have been removed in https://github.com/matrix-org/matrix-appservice-bridge/pull/389.


tadzik (approved):
I'd mention Matrix Spec v1.2 in there (is that the one it landed in?) to be more precise, otherwise looks good.

PR #401, opened 2022-05-18
Remove JS-SDK

### LINKED ISSUE #375: Kill off the js-sdk entirely
We still keep it around for legacy reasons, but there isn't really a reason to. We should just kill it and bump the major version.

Half-Shot:
Fixes #375 

We've been warning about the JS-SDK going away for some time, so this PR just removes it (and associated components).


src/contentRepo.ts
```diff
@@ -0,0 +1,37 @@
...
+ * @param {Number} height The desired height of the thumbnail.
+ * @param resizeMethod The thumbnail resize method to use, either
+ * "crop" or "scale".
+ * @param allowDirectLinks If true, return any non-mxc URLs
+ * directly. Fetching such URLs will leak information about the user to
+ * anyone they share a room with. If false, will return the emptry string
+ * for such URLs.
+ * @return The complete URL to the content. May be empty string if mxc is not a string.
+ */
+function getHttpUriForMxc(baseUrl: string, mxc: string, width?: number, height?: number,
```
tadzik, line 17: Is this borrowed from js/bot-sdk, or written anew? If borrowed, I'd permalink the original source in a comment in case it needs longer than we're planning it to and needs some upstream changes backported.

tadzik (approved):
Looks sensible to me!

src/contentRepo.ts
Half-Shot: It's a weird one, the signature is borrowed but the implementation is written anew. What's best do you think?

tadzik:
Ah, in that case that just looks like sound code to me :) I suspected 
you might have taken it from somewhere, in which case a reference would 
be nice to have. If it's brand new code then I have no complaints :)

PR #405, opened 2022-05-23
Modify index update script to writeback

abbyck:
Also removing the directory `v` which was created by the broken action


Half-Shot (approved):
Oops, thank you!

PR #407, opened 2022-05-23
Update docs for 4.0.0 and 4.0.1

abbyck:
Generated locally as workflow won't be triggered without a release and won't cover old package versions


Half-Shot (approved):
Doh, thank you

Half-Shot (approved):

PR #408, opened 2022-05-23
Add Service Room component

Half-Shot:
This component allows bridges to report issues with their configuration, service outages or other high level issues to a Matrix room which can then be used by other services to report issues to end users. This is intended strictly for server admins.

### DIFF
--- a/src/components/service-room.ts
+++ b/src/components/service-room.ts
@@ -139,6 +139,7 @@ export class ServiceRoom {
 		if (!serviceNotice || 'resolved' in serviceNotice) {
 			return false;
 		}
+		this.lastNoticeTime.delete(noticeId);
 		await this.client.sendStateEvent(
 			this.opts.roomId,
 			STATE_KEY_TYPE,
--- a/src/components/service-room.ts
+++ b/src/components/service-room.ts
@@ -37,6 +37,7 @@ export interface NotificationEventContent {
 	notice_id: string,
 	metadata: Record<string, undefined>;
 	severity: ServiceNotificationServerity;
+	"org.matrix.msc1767.text": string,
 }
 
 interface ResolvedEventContent {
@@ -96,6 +97,7 @@ export class ServiceRoom {
 			notice_id: noticeId,
 			metadata: this.opts.metadata,
 			code,
+			"org.matrix.msc1767.text": `Notice (severity: ${severity}): ${message}`
 		};
 		await this.client.sendStateEvent(
 			this.opts.roomId,
--- a/src/components/service-room.ts
+++ b/src/components/service-room.ts
@@ -25,9 +25,25 @@ export enum ServiceNotificationServerity {
 	Critical = "critical"
 }
 
+
 export interface ServiceRoomOpts {
+	/**
+	 * The roomId to send notices to.
+	 */
 	roomId: string;
+	/**
+	 * The minimum time allowed before
+	 * a new notice with the same ID can be sent (to avoid room spam).
+	 * Defaults to a hour.
+	 */
+	minimumUpdatePeriodMs?: number;
+	/**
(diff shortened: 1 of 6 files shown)


src/components/service-room.ts
```diff
@@ -0,0 +1,167 @@
...
+	/**
+	 * The roomId to send notices to.
+	 */
+	roomId: string;
+	/**
+	 * The minimum time allowed before
+	 * a new notice with the same ID can be sent (to avoid room spam).
+	 * Defaults to a hour.
+	 */
+	minimumUpdatePeriodMs?: number;
```
AndrewFerr, line 40: How about making this non-optional and setting it to a default value in the constructor?

This would also ensure that callers won't have to handle falling back to a default value when checking this.

src/components/service-room.ts
```diff
@@ -0,0 +1,167 @@
...
+	/**
+	 * The prefix to use in state keys to uniquely namespace the bridge.
+	 */
+	bridgeStateKeyPrefix: string;
+	/**
+	 * Any metadata to be included in all notice events.
+	 */
+	metadata: Record<string, unknown>
```
AndrewFerr, line 48: Would it help to define a type alias for this?  It could be shared by both this and `NotificationEventContent.metadata`.

src/components/service-room.ts
```diff
@@ -0,0 +1,167 @@
...
+	metadata: Record<string, unknown>;
+	severity: ServiceNotificationServerity;
+	"org.matrix.msc1767.text": string,
+}
+
+interface ResolvedEventContent {
+	resolved: boolean;
+}
+
+const STATE_KEY_TYPE = "org.matrix.service-notice";
```
AndrewFerr, line 65: Should this be named `EVENT_TYPE` instead?

src/components/service-room.ts
```diff
@@ -0,0 +1,167 @@
...
+			return this.client.getRoomStateEvent(
+				this.opts.roomId,
+				STATE_KEY_TYPE,
+				this.getStateKey(noticeId),
+			);
+		}
+		catch (ex) {
+            if (ex.body.errcode !== "M_NOT_FOUND") {
+                throw ex;
+            }
```
AndrewFerr, line 104: Spaces -> tabs

src/components/service-room.ts
```diff
@@ -0,0 +1,167 @@
...
+	public async getServiceNotification(noticeId: string): Promise<NotificationEventContent|ResolvedEventContent|null> {
+		try {
+			return this.client.getRoomStateEvent(
+				this.opts.roomId,
+				STATE_KEY_TYPE,
+				this.getStateKey(noticeId),
+			);
+		}
+		catch (ex) {
+            if (ex.body.errcode !== "M_NOT_FOUND") {
```
AndrewFerr, line 102: Can `getRoomStateEvent` ever throw an object without a `body`?

src/components/service-room.ts
```diff
@@ -0,0 +1,167 @@
...
+	/**
+	 * Resolve a previous notice to say that the specific issue has been resolved.
+	 * @param noticeId The noticeId to resolve.
+	 * @returns `true` if the notice exists and was resolved,
+	 * 			`false` if the notice did not exist or was already resolved.
+	 */
+	public async clearServiceNotice(noticeId: string): Promise<boolean> {
+		const serviceNotice = await this.getServiceNotification(noticeId);
+		if (!serviceNotice || 'resolved' in serviceNotice) {
+			return false;
```
AndrewFerr, line 152: What about if `serviceNotice.resolved` is false?

(The code currently prevents that from ever happening, but maybe a future change will allow it.)

AndrewFerr (changes requested):

Half-Shot:

PR #409, opened 2022-05-25
Remove option to disable authentication of the 3pid endpoints

### LINKED ISSUE #76: Third Party Endpoints don't check access tokens
A homeserver is able to request third party lookup information without requiring an `access_token`.

This will break existing stuff, also a spec PR should be made first.

Half-Shot:
Fixes #76 

As of https://github.com/matrix-org/synapse/pull/12746, Synapse now correctly sends us an access token when handling third party requests. There is no longer any need leave these endpoints unauthenticated.

### DIFF
--- a/src/bridge.ts
+++ b/src/bridge.ts
@@ -825,7 +825,6 @@ export class Bridge {
             this.addAppServicePath({
                 method: "GET",
                 path: "/_matrix/app/:version(v1|unstable)/thirdparty/protocol/:protocol",
-                checkToken: true,
                 handler: async (req, res) => {
                     const protocol = req.params.protocol;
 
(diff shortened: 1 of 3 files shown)


src/bridge.ts
```diff
@@ -828,7 +825,7 @@ export class Bridge {
             this.addAppServicePath({
                 method: "GET",
                 path: "/_matrix/app/:version(v1|unstable)/thirdparty/protocol/:protocol",
-                checkToken: this.opts.authenticateThirdpartyEndpoints,
+                checkToken: true,
```
AndrewFerr: Nitpick: unless there's a good reason for this to specifically set `checkToken`, just get rid of it, like all of the other calls to `this.addAppServicePath` did.

AndrewFerr (approved):

AndrewFerr:
Every CI failure is triggered by the same issue in `src/components/prometheusmetrics.ts`.  Given this PR doesn't touch that file, this should be fine to merge, right?

Half-Shot:
> Every CI failure is triggered by the same issue in `src/components/prometheusmetrics.ts`. Given this PR doesn't touch that file, this should be fine to merge, right?

Ah, but I had failed to update that file after changing the signature!

PR #410, opened 2022-06-06
Request logging refactor

Half-Shot:
Depends on #412 (Based on `hs/logging-refactor`, change branch before merge)

This PR is quite an invasive one, but since we're planning a breaking release there isn't a better time to try. The intent here is to improve several components of the library:

Taking inspiration from Synapse and the teams bridge, this improves our Request components:
 - Request now extends the Logger class, so that messages are logged with a request ID.
 - Matrix events now use the MatrixRequest class types when handled through the `controller`, which should hopefully improve types.

The hope is that bridges further extend the request type for remote requests too, so that remote requests also gain their own IDs.


Half-Shot:

PR #412, opened 2022-06-08
Logging system refactor

Half-Shot:
Split out from #410 

Our current logging system is hard to use, and also difficult to override. As such we've ended up building our own solutions in the Discord bridges, IRC bridge and the hookshot bridge. This PR attempts to pull in the hookshot way of doing logging into this bridge without trying to break the API entirely. 

I've retained our existing log rotation transport feature, but I'm not wholly convinced it's something we want long term. Log rotation is best left to other services (KISS principle applies).

For reference https://github.com/matrix-org/matrix-hookshot/blob/main/src/LogWrapper.ts

---

The PR rotted enough that trying to merge some changes in broke the world, so I have restarted again on a single PR.

### DIFF
--- a/src/components/logging.ts
+++ b/src/components/logging.ts
@@ -1,5 +1,5 @@
 /*
-Copyright 2020 The Matrix.org Foundation C.I.C.
+Copyright 2022 The Matrix.org Foundation C.I.C.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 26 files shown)


turt2live:
(in case it was missed: the bot-sdk has a `LogService` which it uses internally and can be used externally too)

Half-Shot:
>(in case it was missed: the bot-sdk has a LogService which it uses internally and can be used externally too)

Yep, the current problem is that it lacks the ability to set freeform metadata on the Logging object which is something we're going to end up needing for the Request logging we have planned. Given the time-to-merge delays on matrix-bot-sdk, it's not something we're going to be able to depend on easily, hence running out own solution. We still wrap the bot-sdk logger so that the logs get piped nicely into one place.

src/components/logging.ts
```diff
@@ -12,221 +12,372 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+export interface LoggingOptsFile extends LoggingOpts {
+    /**
+     * An object mapping a file name to a logging level. The file will contain
+     * all logs for that level inclusive up to the highest level. (`info` will contain `warn`, `error` etc)
+     * Use `%DATE%` to set the date of the file within the string.
+     * Use the `fileDatePattern` to set the date format.
+     * @example {"info-%DATE%.log": "info"}
+     */
+    files: {
+        [filename: string]: LogLevel,
```
jaller94, line 105: /me starts naming his log files `__proto__`. :foot::gun:

src/components/logging.ts
```diff
@@ -12,221 +12,372 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+ * @returns True is the message is noise, or false otherwise.
+ */
+function isMessageNoise(messageOrObjects: MsgType[]) {
+	return !!messageOrObjects.find(messageOrObject => {
+		if (typeof messageOrObject !== "object") {
+			return false;
+		}
+
+		const possibleError = messageOrObject as {
+            error?: string, body?: { error?: string, errcode?: string}, errcode?: string
```
jaller94, line 37: Mix of tabs and space indentation. :-1:

src/components/logging.ts
```diff
@@ -12,221 +12,372 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+
+type MsgType = string|Error|any|{error?: string};
+
+/**
+ * Tries to filter out noise from the bot-sdk.
+ * @param messageOrObjects A list of values being logged.
+ * @returns True is the message is noise, or false otherwise.
+ */
+function isMessageNoise(messageOrObjects: MsgType[]) {
+	return !!messageOrObjects.find(messageOrObject => {
```
jaller94:
```suggestion
	return messageOrObjects.some(messageOrObject => {
```

There's ~an app~ a method for that. :green_apple:

jaller94: I was confused whether this could be a single message (not an array).
```suggestion
function isMessageNoise(messagesOrObjects: MsgType[]) {
	return !!messagesOrObjects.find(messageOrObject => {
```

Half-Shot: fancy!

src/components/logging.ts
Half-Shot: AH! Yes!

jaller94 (changes requested):
Looks good, but the `fileName` as Object key should get resolved before we merge it.

src/components/logging.ts
Half-Shot: I've ended up naming the type and the variable to make more sense.

src/components/logging.ts
Half-Shot: Eurgh, wish the linter picked up on this.

src/components/logging.ts
Half-Shot: This is difficult, it's an interface to the logger config so we can't make changes lightly. This is also only really configured by administrators and ultimately comes from a yaml file somewhere. I feel like we're relatively safe from `__proto__` corruption here:

1. You'd have to as an admin explicitly put this key in, which feels like something you'd never really do.
2. You'd only be harming your own software in doing so.

It does feel a bit strange to ask people to submit a map as a parameter to a function if it needs to be parsed from an object first during YAML parsing, so I'm hesistant to say we should change this.

jaller94 (approved):
Looks good to me.

PR #414, opened 2022-07-14
Account for the port specifier at the end of server names

Gnuxie:
I guess this also works for IPv6 literals.
https://spec.matrix.org/v1.3/appendices/#server-name

### DIFF
--- a/changelog.d/414.bugfix
+++ b/changelog.d/414.bugfix
@@ -1,0 +1,1 @@
+Matrix users on a server that has an explicit port specifier in the server name will now be supported.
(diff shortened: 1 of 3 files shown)


Half-Shot:
Oops, thank you. Could I request a test or two in https://github.com/matrix-org/matrix-appservice-bridge/blob/develop/spec/unit/matrix-user.spec.js?

Gnuxie:
okie, is this what you had in mind?

Half-Shot (approved):

PR #415, opened 2022-07-15
Update matrix-bot-sdk and other packages

Half-Shot:
This updates the bot-sdk to 0.6.0, and other packages to their latest versions (where possible). We've also had to deprecate uploading files directly from a ReadStream as converting RS into Buffer isn't possible on Node 14 directly, and writing the code to do so felt out of scope for this library.


Half-Shot:
Broken due to https://github.com/turt2live/matrix-bot-sdk/issues/247

tadzik (approved):
Looks good – though I wonder where the ReadStream deprecation is going to bite us. Certainly makes sense to not do it here though.

Issue #421, opened 2022-07-20
onUserQuery should not swallow exception


tak-hntlabs:
**Describe the bug**
In bridge.ts, the method Bridge.onUserQuery swallows exceptions and logs a message. Consequently AppService.onGetUsers sends back a http 200 to the dendrite server. The server assumes that the user query succeeded. Eventually it ran into some other issues.

**Expected behavior**
Propose removing try..catch in Bridge.onUserQuery. Let AppService catch the exception and send back a 400 status. The dendrite / synapse server can handle it as appropriate.

PR #422, opened 2022-07-20
remove try catch from onUserQuery

tak-hntlabs:
Proposed fix for issue: https://github.com/matrix-org/matrix-appservice-bridge/issues/421

Related to this issue: https://github.com/matrix-org/matrix-appservice-node/issues/55


tak-hntlabs:
Switched over to the matrix-bot-sdk.  Not needing this any more.

PR #428, opened 2022-07-26
Add signoff script

Half-Shot:
This does a simple check to ensure that the PR contains a sign off if the contributor isn't a member of the matrix-org project.


tadzik (approved):
Looks correct to me. Is that some standard script, maybe something we can share between repos?

Half-Shot:
I'm hoping we can do a simple import between repos.

PR #430, opened 2022-07-29
Document platform support

Half-Shot:
This document aims to cover how the versioning system works for bridges, what node versions we support, and what target platforms we support. The intention is for this to be linked to by the other bridges and for us to have a common baseline across projects.

### DIFF
--- a/docs/platforms.md
+++ b/docs/platforms.md
@@ -11,7 +11,10 @@ All bridge application and library projects aim to follow the [Semantic Versioni
 Bridge projects will always support the **Active LTS** release and the **Current** release of Node.JS. When a
 Node.JS version falls from Active to Maintenance, we will migrate projects to the next Active release over time.
 
-At time of writing (July 2022) we support 16.X and 18.X as they are the Active and Current releases of Node.
+For users who are on distributions packaging only Maintenance versions of Node.JS, we'd suggest either using Docker
+or an alternative Node.JS source.
+
+At time of writing (July 2022) we support 16.X and 18.X as they are the Active and Current releases of Node.JS.
 
 Bridge projects do not support odd-versioned Node.JS releases, as these are short lived non-LTS versions and are
 difficult to support since they have a 6 month shelf life.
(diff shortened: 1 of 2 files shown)


docs/platforms.md
```diff
@@ -0,0 +1,41 @@
...
+We do not currently have a LTS process for bridge projects. The latest released version of the bridge is
+what we aim to support. Versions a few points behind latest might be eligible for some support, but the
+advice is always to upgrade as soon as possible.
+
+### Platforms
+
+In terms of installation platforms, we support as a baseline:
+
+- Installation onto a host directly (via git) on Windows, Mac, Linux.
+- Docker (x86_64) on Windows, Mac, Linux.
```
tadzik: I think we should specify architecture in both cases, if applicable - unless it only matters in case of docker?

Maybe also add a note saying "we test things on Debian Stale/Ubuntu LTS/whatever, but we'll help you anyway with whatever distro you have" (which I think we can safely commit to until proven otherwise).

docs/platforms.md
```diff
@@ -0,0 +1,41 @@
...
+
+Support for other architectures is something we aim to do, for instance ARM but at the moment
+this is best-effort.
+
+In a few cases, bridges also have a maintained version on npm but this is not universally true.
```
tadzik: Do they have unmaintained version on npm, or no versions at all? If it's the former then we should probably drop them, before we forget to update them with security fixes or something :)

Half-Shot: I believe we always update the IRC bridge, but we don't provide it for the others.

docs/platforms.md
Half-Shot: Yeah, might as well say we have tested on x86_64 Debian ( & derivatives) but we expect any Linux distro to work. In addition, other architectures will *probably* be fine.

docs/platforms.md
```diff
@@ -0,0 +1,41 @@
...
+This document outlines the supported platforms for matrix.org bridge projects.
+
+### Bridge versioning
+
+All bridge application and library projects aim to follow the [Semantic Versioning](https://semver.org/) system.
+
+### Node version support
+
+Bridge projects will always support the **Active LTS** release and the **Current** release of Node.JS. When a
+Node.JS version falls from Active to Maintenance, we will migrate projects to the next Active release over time.
```
AndrewFerr, line 12: Do we provide no guarantees of supporting the Maintenance LTS release?  Would that make support on "extra-stable" distros (i.e. Debian) difficult?

Half-Shot: Debian is very hard to support because the stable versions pack out-of-LTS releases. We're never going to be able to support that. 

I'd *like* to support maintenance versions a bit longer, but the trouble is upstream projects like bot-sdk make that a bit difficult (it has a hard dep on Node 16+).

docs/platforms.md
Half-Shot: >I think we should specify architecture in both cases, if applicable - unless it only matters in case of docker?

Ah, should clarify. We build x84_64 images for Docker automatically, but we don't specifically support an arch so others should be fine.

Half-Shot: Hopefully better?

docs/platforms.md
Half-Shot: Decided to drop this statement in the end, given we only do it for IRC.

docs/platforms.md
```diff
@@ -0,0 +1,44 @@
+# Platforms
+
+This document outlines the supported platforms for matrix.org bridge projects.
```
justinbot, line 3: Is this doc going to be copied or linked to from the downstream projects?

docs/platforms.md
tadzik: LGTM :)

docs/platforms.md
Half-Shot: That's the plan yep.

AndrewFerr (approved):
LGTM!

justinbot (approved):

tadzik (approved):

PR #434, opened 2022-08-12
Ensure onUserQuery errors are passed through to the HTTP handler

Half-Shot:
This essentially incorporates the feature in https://github.com/matrix-org/matrix-appservice-node/releases/tag/1.1.0


jaller94 (approved):
Looking good.

PR #435, opened 2022-08-12
Add checkToken back to addAppServicePath and ensure metrics doesn't require a token

Half-Shot:
This was broken in a previous release where we suddenly enabled authentication forcefully across all our endpoints, which broke bridges that used it for unauthenticated metrics access.

### DIFF
--- a/src/bridge.ts
+++ b/src/bridge.ts
@@ -960,23 +960,23 @@ export class Bridge {
      * @param opts Named options
      * @param opts.method The HTTP method name.
      * @param opts.path Path to the endpoint.
-     * @param opts.checkToken Should the token be automatically checked. Defaults to true.
+     * @param opts.authenticate Should the token be automatically checked. Defaults to true.
      * @param opts.handler Function to handle requests
      * to this endpoint.
      */
     public addAppServicePath(opts: {
         method: "GET"|"PUT"|"POST"|"DELETE",
         path: string,
-        checkToken?: boolean,
+        authenticate: boolean,
         handler: (req: ExRequest, respose: ExResponse, next: NextFunction) => void,
     }): void {
         if (!this.appservice) {
             throw Error('Cannot call addAppServicePath before calling .run()');
         }
         const app: Application = this.appservice.expressApp;
-
+        const authenticate = opts.authenticate ?? true;
         app[opts.method.toLowerCase() as "get"|"put"|"post"|"delete"](opts.path, (req, res, ...args) => {
-            if (opts.checkToken === false || !this.requestCheckToken(req)) {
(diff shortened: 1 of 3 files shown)


src/bridge.ts
```diff
@@ -967,6 +967,7 @@ export class Bridge {
     public addAppServicePath(opts: {
         method: "GET"|"PUT"|"POST"|"DELETE",
         path: string,
+        checkToken?: boolean,
```
jaller94:
```suggestion
        checkToken: boolean = true,
```

Given the `checkToken === false`, this defaults to `true`.
Let's make that explicit in the signature.

I'm a fan of defaulting to `false`. Have you considered `ignoreToken` as a variable name?

jaller94 (changes requested):

src/bridge.ts
Half-Shot: Hm, maybe we should just have `authenticate` to be properly explicit. `checkToken` was used as the legacy variable name, but since that was removed in 5.0.0 and we have typings, I don't think we need to follow it.

src/bridge.ts
```diff
@@ -958,24 +958,25 @@ export class Bridge {
...
         path: string,
+        authenticate?: boolean,
         handler: (req: ExRequest, respose: ExResponse, next: NextFunction) => void,
     }): void {
         if (!this.appservice) {
             throw Error('Cannot call addAppServicePath before calling .run()');
         }
         const app: Application = this.appservice.expressApp;
-
+        const authenticate = opts.authenticate ?? true;
```
jaller94, line 977:
```suggestion
        authenticate = opts.authenticate ?? true;
```

Is re-declaring a parameter on the top level of a function valid in JavaScript?
I'm getting `SyntaxError: Identifier 'hi' has already been declared` in a local test.

Is the linter dead?

jaller94 (changes requested):

jaller94 (approved):
Nevermind, the parameter is `opts.authenticate` not `authenticate`.

src/bridge.ts
Half-Shot: Should be fine? It's an opts object, not a parameter. (Which is why I can't do defaults)

src/bridge.ts
Half-Shot: Oh I should have mentioned that I couldn't do a explicit signature value because it's an opts object type. I have however been explicit in the code below.

On the naming, on the whole I think that having a positive variable name like `authenticate` is probably a better look than `ignoreToken` because negative variable names are usually bad. I do concede that having default-to-true variables are a bit spooky, I hope people read the docstring :)

PR #437, opened 2022-09-13
Improve package version fetching reliability

### LINKED ISSUE #436: getBridgeVersion isn't reliable
It rarely finds the package.json of the calling package, which is obviously a hard problem but we could probably make a stronger attempt.

Half-Shot:
Fixes #436 

This code now looks for the package.json adjacent to the main module script, and iterates backwards. This is only done once on first fetch, so should be fairly performant.


src/utils/package-info.ts
```diff
@@ -1,5 +1,11 @@
-import { join, resolve } from "path";
-let BridgeVersion: string;
+import { join, dirname } from "path";
+import { statSync } from "fs";
+import Logging from "../components/logging";
+
+const log = Logging.get("PackageInfo");
+
+// This may be defined if the script is run via NPM: https://docs.npmjs.com/cli/v8/using-npm/scripts#packagejson-vars
+let BridgeVersion: string|undefined = process.env.npm_package_version;
```
justinbot, line 9: Could this default to `"unknown"` and then the type can just be `string`?
(To simplify `getBridgeVersion` a bit)

src/utils/package-info.ts
```diff
@@ -11,23 +17,54 @@ export function setBridgeVersion(version: string): void {
     BridgeVersion = version;
 }
 
+/**
+ * Try to determine the path of the `package.json` file for the current
+ * running module. Iterates through parent directories of `require.main.filename`
+ * until it finds a package.json. This **may** result in false positives.
+ * @returns The path to a package.json file, or undefined if one could not be found.
+ */
+export function identifyPackageFile(): string|undefined {
```
justinbot, line 27: There's a (old but very popular) package which does this: https://www.npmjs.com/package/pkginfo

Here's how it finds the `package.json` https://github.com/indexzero/node-pkginfo/blob/f4e90ac77034ee9eb3dd7aafc56657d7376795a8/lib/pkginfo.js#L89

justinbot (approved):
A couple suggestions but otherwise looks good

src/utils/package-info.ts
Half-Shot: Yeah, I wonder if we should just be using this :thinking:. Saves us having to maintain this code.

src/utils/package-info.ts
Half-Shot: Not easily, it's designed so that it's explicitly undefined if we've not found a version yet. `getBridgeVersion` will either use it if defined, or fetch the version and then set it.

PR #440, opened 2022-10-05
Fix initialise typo 

Half-Shot:


jesopo (approved):

AndrewFerr:
Note that since `Bridge.initialise` is an exported function, this is an API-breaking change.

PR #441, opened 2022-10-05
Use MatrixError

Half-Shot:
Successor to #360, but more up to date

### DIFF
--- a/spec/unit/intent.spec.js
+++ b/spec/unit/intent.spec.js
@@ -1,6 +1,6 @@
 const { Intent } = require("../..");
-
-const matrixError = (errcode, error) => Promise.reject({body: {errcode, error}});
+const { MatrixError } = require('matrix-bot-sdk');
+const matrixError = async (errcode, error) => { throw new MatrixError({errcode, error}, 500)};
 
 describe("Intent", function() {
     let intent, botIntent, client, botClient, underlyingClient;
--- a/src/components/membership-queue.ts
+++ b/src/components/membership-queue.ts
@@ -2,6 +2,7 @@ import { Bridge } from "../bridge";
 import { Logger } from ".."
 import PQueue from "p-queue";
 import { Counter, Gauge } from "prom-client";
+import { MatrixError } from "matrix-bot-sdk";
 
 const log = new Logger("MembershipQueue");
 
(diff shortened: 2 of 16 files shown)


src/components/room-upgrade-handler.ts
```diff
@@ -107,7 +108,8 @@ export class RoomUpgradeHandler {
             return true;
         }
         catch (ex) {
-            if (ex.body?.errcode === "M_FORBIDDEN") {
+            console.log("THE EX", ex);
```
jaller94: This was left from debugging, I guess?.
Otherwise, using a `log` library and a proper description would be much appreciated. :)
```suggestion
```

spec/unit/intent.spec.js
```diff
@@ -1,6 +1,6 @@
 const { Intent } = require("../..");
-
-const matrixError = (errcode, error) => Promise.reject({body: {errcode, error}});
+const { MatrixError } = require('matrix-bot-sdk');
+const matrixError = async (errcode, error) => { throw new MatrixError({errcode, error}, 500)};
```
jaller94, line 3: When to use the second parameter of `MatrixError.constructor` and when not?

jaller94 (approved):
Looks good.

src/components/room-upgrade-handler.ts
Half-Shot: Definitely debugging. Definitely not helpful to anyone else :)

spec/unit/intent.spec.js
Half-Shot: Always really, this is just laziness...

AndrewFerr:
Turning on `useUnknownInCatchVariables` in the tsconfig exposes some more places where `MatrixError` can/should be used instead.  I'll get on it because the lack of typed errors has been bugging me more and more lately & I'd love to have it in.

src/components/membership-queue.ts
```diff
@@ -279,15 +280,17 @@ export class MembershipQueue {
...
+            return !(
+                attempts === this.opts.maxAttempts ||
+                // Forbidden
+                ex.errcode === "M_FORBIDDEN" ||
+                ex.statusCode === 403 ||
+                // Not found
+                ex.statusCode === 404
+            )
+        }
+        return false;
```
AndrewFerr: Should this return `true` instead?  If `ex` isn't a `MatrixError`, the original code can resolve to `!(undefined === <something>)` -> `!false` -> `true`

src/components/stores/event-store.ts
```diff
@@ -1,43 +0,0 @@
-import { StoredEvent } from "../../models/events/event";
```
AndrewFerr: I may rebase this PR to split out these file removals (which have since been committed as 7018468).

AndrewFerr (dismissed):
I'll see how far I can get with enforcing `useUnknownInCatchVariables`.

PR #442, opened 2022-10-06
Add PostgresStore abstraction

Half-Shot:
This semi-ports the IRC bridge implementation of a Postgres Datastore to this library. There are some key differences though:

- It's a very cut back abstract class that only deals with schema updates, for now.
- It deliberately doesn't try to implement any of our datastore classes for the moment, leaving that exercise to the bridge developer.
- We use the https://www.npmjs.com/package/postgres library, which has a (IMO) nicer syntax and natural support for TypeScript.
- Schema updates are now more typesafe than before.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -58,5 +58,5 @@ jobs:
       - run: yarn --frozen-lockfile
       - run: yarn test
         env:
-          BRIDGE_TEST_PGDB: "ircbridge_integtest"
+          BRIDGE_TEST_PGDB: "bridge_integtest"
           BRIDGE_TEST_PGURL: "postgresql://postgres_user:postgres_password@postgres"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -55,8 +55,8 @@ jobs:
         uses: actions/setup-node@v1
         with:
           node-version: 18
-      - run: npm ci
-      - run: npm run-script test
+      - run: yarn --frozen-lockfile
+      - run: yarn test
         env:
           BRIDGE_TEST_PGDB: "ircbridge_integtest"
           BRIDGE_TEST_PGURL: "postgresql://postgres_user:postgres_password@postgres"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,3 +31,32 @@ jobs:
           node-version: "${{ matrix.node-version }}"
       - run: yarn --frozen-lockfile
       - run: yarn build && yarn test
+  test-postgres:
+    runs-on: ubuntu-20.04
+    container: node:18
+    services:
+      postgres:
+        image: postgres:latest
+        env:
+          POSTGRES_DB: postgres
+          POSTGRES_PASSWORD: postgres_password
+          POSTGRES_PORT: 5432
+          POSTGRES_USER: postgres_user
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+    steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: 18
+      - run: npm ci
+      - run: npm run-script test
+        env:
+          BRIDGE_TEST_PGDB: "ircbridge_integtest"
+          BRIDGE_TEST_PGURL: "postgresql://postgres_user:postgres_password@postgres"
--- a/spec/helpers/postgres-helper.ts
+++ b/spec/helpers/postgres-helper.ts
@@ -1,0 +1,21 @@
+import postgres, { Sql } from 'postgres';
+
+let pgClient: Sql;
+
+export function isPostgresTestingEnabled() {
+    return !!process.env.BRIDGE_TEST_PGURL;
+}
+
+export function initPostgres() {
+        // Setup postgres for the whole process.
+        pgClient = postgres(`${process.env.BRIDGE_TEST_PGURL}/postgres`);
+        process.on("beforeExit", async () => {
+            pgClient.end();
+        })
+}
+
+export async function getPgDatabase() {
+    const pgDb = `${process.env.BRIDGE_TEST_PGDB}_${process.hrtime().join("_")}`;
+    await pgClient`CREATE DATABASE ${pgClient(pgDb)}`;
+    return `${process.env.BRIDGE_TEST_PGURL}/${pgDb}`;
+}
--- a/spec/helpers/postgres-helper.ts
+++ b/spec/helpers/postgres-helper.ts
@@ -7,11 +7,11 @@ export function isPostgresTestingEnabled() {
 }
 
 export function initPostgres() {
-        // Setup postgres for the whole process.
-        pgClient = postgres(`${process.env.BRIDGE_TEST_PGURL}/postgres`);
-        process.on("beforeExit", async () => {
-            pgClient.end();
-        })
+    // Setup postgres for the whole process.
+    pgClient = postgres(`${process.env.BRIDGE_TEST_PGURL}/postgres`);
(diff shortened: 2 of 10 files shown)


spec/helpers/postgres-helper.ts
```diff
@@ -0,0 +1,21 @@
...
+export function isPostgresTestingEnabled() {
+    return !!process.env.BRIDGE_TEST_PGURL;
+}
+
+export function initPostgres() {
+        // Setup postgres for the whole process.
+        pgClient = postgres(`${process.env.BRIDGE_TEST_PGURL}/postgres`);
+        process.on("beforeExit", async () => {
+            pgClient.end();
+        })
```
AndrewFerr: Nitpick: shift left by one intent level

spec/helpers/postgres-helper.ts
```diff
@@ -0,0 +1,21 @@
...
+}
+
+export async function getPgDatabase() {
+    const pgDb = `${process.env.BRIDGE_TEST_PGDB}_${process.hrtime().join("_")}`;
+    await pgClient`CREATE DATABASE ${pgClient(pgDb)}`;
+    return `${process.env.BRIDGE_TEST_PGURL}/${pgDb}`;
+}
```
AndrewFerr, line 21: The lack of a newline fires an eslint warning, which is IMO frivolous for a test file.  And apparently `spec/`'s eslint config file is broken, so here's one that works if you want it:
```
{
  "rules": {
    "strict": "off",
    "no-var": "off",
    "eol-last": "off"
  }
}
```

spec/integ/postgres.spec.ts
```diff
@@ -0,0 +1,26 @@
+import { PostgresStore } from "../../src";
+import { getPgDatabase, initPostgres, isPostgresTestingEnabled } from "../helpers/postgres-helper";
+
+
+// Only run the tests if we've enabled postgres testing.
+let descr = isPostgresTestingEnabled() ? describe : xdescribe;
```
AndrewFerr: Can be `const`

.github/workflows/test.yml
```diff
@@ -30,4 +30,33 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: 18
+      - run: yarn --frozen-lockfile
+      - run: yarn test
+        env:
+          BRIDGE_TEST_PGDB: "ircbridge_integtest"
```
AndrewFerr: Not a big deal, but maybe take out `irc` from this to keep it generic

src/components/stores/postgres-store.ts
```diff
@@ -0,0 +1,153 @@
+import postgres, { PostgresError, PostgresType } from 'postgres';
```
AndrewFerr, line 1: Any reason why `Sql` is imported by name in tests but not here?

src/components/stores/postgres-store.ts
```diff
@@ -0,0 +1,153 @@
...
+    /**
+     * Should the schema table be automatically created (the v0 schema effectively)
+     */
+    autocreateSchemaTable?: boolean;
+}
+
+export type SchemaUpdateFunction = (sql: postgres.Sql) => void;
+
+/**
+ * PostgreSQL datastore abstraction which can be inherited by a specalised bridge class.
```
AndrewFerr: specalised -> specialised

src/components/stores/postgres-store.ts
```diff
@@ -0,0 +1,153 @@
...
+ *   }
+ * }
+ *
+ * // Which can then be used by doing
+ * const store = new MyBridgeStore("postgresql://postgres_user:postgres_password@postgres");
+ * store.ensureSchema();
+ * const data = await store.getData();
+ */
+export abstract class PostgresStore {
+    public static readonly LATEST_SCHEMA = 9;
```
AndrewFerr: Is this vestigial?

src/components/stores/postgres-store.ts
```diff
@@ -0,0 +1,153 @@
...
+            // We aren't autocreating the schema table, so assume schema 0.
+            currentVersion = 0;
+        }
+
+        // Zero-indexed, so schema 1 would be in slot 0.
+        while (this.schemas[currentVersion]) {
+            log.info(`Updating schema to v${currentVersion + 1}`);
+            const runSchema = this.schemas[currentVersion];
+            try {
+                await runSchema(this.sql);
```
AndrewFerr, line 99: Are the schema update functions supposed to be async?  Their typehint makes them synchronous, since it says they return `void` instead of `Promise<void>`

src/components/stores/postgres-store.ts
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    /**
+     * Get the current schema version.
+     * @returns The current schema version, or `-1` if no schema table is found.
+     */
+    protected async getSchemaVersion(): Promise<number> {
+        try {
+            const result = await this.sql<{version: number}[]>`SELECT version FROM SCHEMA;`;
+            return result?.[0]?.version;
```
AndrewFerr: Does this need to check for `undefined`?  Or does it get caught by the `PostgresError`?
```suggestion
            return result?.[0]?.version ?? -1;
```

spec/integ/postgres.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+        initPostgres();
+    })
+
+    it('can construct a simple database from schema', async () => {
+        store = new TestPostgresStore([], {
+            url: await getPgDatabase(),
+        });
+        await store.ensureSchema();
+    });
+
```
AndrewFerr, line 22: This would benefit from a test that runs schema upgrades.  Here's a simple one that I've confirmed passes:
```suggestion
    it('can run schema upgrades', async () => {
        store = new TestPostgresStore([
            sql => sql.begin(s => [
                s`CREATE TABLE v1_users (mxid TEXT UNIQUE NOT NULL);`,
            ]),
            sql => sql.begin(s => [
                s`CREATE TABLE v2_rooms (mxid TEXT UNIQUE NOT NULL);`,
            ]),
        ], {
            autocreateSchemaTable: true,
            url: await getPgDatabase(),
        });
        await store.ensureSchema();
    });

```

AndrewFerr (changes requested):

spec/helpers/postgres-helper.ts
Half-Shot: Ah, derp. Thanks!

.github/workflows/test.yml
Half-Shot: Oh yeah, copy paste bug.

src/components/stores/postgres-store.ts
Half-Shot: We're importing Sql in tests? I don't see it.

src/components/stores/postgres-store.ts
Half-Shot: Another borrow from the IRC bridge.

src/components/stores/postgres-store.ts
Half-Shot: Good catch!

src/components/stores/postgres-store.ts
Half-Shot: Ah, this is probably safe tbh.

spec/integ/postgres.spec.ts
Half-Shot: That would be helpful :grin: . Thanks!

src/components/stores/postgres-store.ts
AndrewFerr: I just meant this line in `postgres-helper.ts`:
```ts
import postgres, { Sql } from 'postgres';
```
It's just a convenience renaming, so it's not important to do it here too.

AndrewFerr (approved):

PR #443, opened 2022-10-12
Add allowedIpRanges to ProvisionApiOpts

benbz:
Similar to Synapse's [ip_range_whitelist](https://matrix-org.github.io/synapse/latest/usage/configuration/config_documentation.html#ip_range_whitelist) add `allowedIpRanges` to `ProvisioningApiOpts` such that `disallowedIpRanges` can be kept simple and clear and if specific sub-ranges need to be allowed they can easily be specified without fraught editing of `disallowedIpRanges`.

I've never done typescript before, so there's significant potential for this to be incorrect but it seems simple and self-contained enough. Best I can tell ordering on `ProvisioningApiOpts` isn't important and that fields can be optional so this should be backwards compatible with existing users of this interface.

### DIFF
--- a/changelog.d/443.feature
+++ b/changelog.d/443.feature
@@ -1,0 +1,1 @@
+Enable allow-listing of specific IP ranges for OpenAPI requests without having to edit disallowedIpRanges.
--- a/src/provisioning/api.ts
+++ b/src/provisioning/api.ts
@@ -61,6 +61,16 @@ export interface ProvisioningApiOpts {
      */
     openIdOverride?: {[serverName: string]: URL},
     /**
+     * Allow these IP ranges to be hit when handling OpenID requests even if they are within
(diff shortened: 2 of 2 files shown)


changelog.d/443.feature
```diff
@@ -0,0 +1 @@
+Enable allow-listing of specific IP ranges for OpenAPI requests without having to edit disallowedIpRanges.
```
Half-Shot, line 1:
```suggestion
Enable allow-listing of specific IP ranges for OpenAPI requests (when using the Provisioner API) without having to edit disallowedIpRanges. This is done via the new `allowedIpRanges` flag. 
```

Half-Shot (approved):
Looks great! Nice Typescript :)

changelog.d/443.feature
benbz: Ah you merged this too quickly for me to accept this suggestion :)

PR #452, opened 2022-11-30
Release helper script and CI

Half-Shot:
This should help people other than me release changes for matrix-appservice-bridge.

This comes in 3 parts:
- The `release.sh` script replaces the old `changelog-release.sh` script. It will now generate a draft changelog, check you are happy with it, then create the correct commits, tags and push the right things.
- A CI that takes the incoming changelog and creates a release automatically, you are then invited to check the draft to make sure it is sane.
- Once that's complete, there is a final CI job to build the npm package and publish it.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,28 @@
+
+on:
+  push:
+    tags: ["[0-9].[0-9].[0-9]"]
+
+jobs:
+  do-release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Get changelog
+        id: extract-changelog
+        run: |
+          RELEASE_BODY=`git diff --minimal --no-color HEAD~ CHANGELOG.md | tail -n +6 | sed -r "s/^([^-+ ]*)[-+ ]/\\1/" | head -n -3`
+          RELEASE_NAME=`echo $RELEASE_BODY | head -n 1`
+          RELEASE_BODY=`echo $RELEASE_BODY | tail -n +3`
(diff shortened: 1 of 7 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,25 @@
+
+on:
+  push:
+    tags: ["testing-[0-9].[0-9].[0-9]"]
```
AndrewFerr: Does the `"testing-` mean that this will only match when the push contains a commit with a tag named "testing-..."?

Half-Shot: Yeah, I apparently left this one in.

scripts/release.sh
```diff
@@ -0,0 +1,41 @@
...
+towncrier build --version $VERSION
+git commit CHANGELOG.md changelog.d/ package.json -m $TAG
+
+echo "Proceeding to generate tags"
+cat draft-release.txt | git tag --force -m - -s $TAG
+rm draft-release.txt
+echo "Generated tag $TAG"
+
+echo "Pushing to origin"
+git push origin $TAG
```
Half-Shot, line 43: This obviously doesn't allow us to do a PR for releases, which may not be desirable. Thoughts on how this could be re-architected to support that are welcome. Alternatively we can forgo the PR.

AndrewFerr: PRs for releases can be a bit of a pain, and are probably redundant since we usually discuss release plans internally before publishing anything.  So I'm in favour of forgoing.

AndrewFerr (approved):
LGTM

PR #453, opened 2022-12-01
Use `once` with process 'beforeExit'.

Gnuxie:
This is because when we call `PostgresStore.destroy`, we put a new task on the event loop. This "revives" the process and so the callback for process 'beforeExit' will be called repeatedly.

Signed-off-by: Gnuxie <gnuxie@element.io>

### DIFF
--- a/changelog.d/453.bugfix
+++ b/changelog.d/453.bugfix
@@ -1,0 +1,1 @@
+The PostgresStore had a beforeExit hook that would itself prevent the process from exiting.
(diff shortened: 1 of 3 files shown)


Half-Shot (approved):
Seems reasonable, is that the only "beforeExit" handler? Do we have any others to fix?

Gnuxie:
Only in `spec/helpers/postgres-helper.ts`, I have changed it too but I'm not sure on the context of its use.

I also don't know why the contribution requirements isn't happy on this PR?

Half-Shot:
>I also don't know why the contribution requirements isn't happy on this PR?

Brackets!

PR #458, opened 2023-02-06
Improve Provisioning API rate limiting

justinbot:
Improves the Provisioning API rate limiting implementation:
* Updates `express-rate-limit`
* Corrects the parameter type to `Partial` rate limit options, so that bridges can pass overrides if necessary
* Changes the default from 100req/6 minutes to 30req/1 minute
    * I think this is more balanced for a default rate limit, allowing a lower burst while having a shorter period before the rate limit resets. In any case, bridges can override this if necessary.

And finally, fixes the rate limit response.
Previously, response headers looked like this:
```
X-RateLimit-Limit: 20
X-RateLimit-Remaining: 18
X-RateLimit-Reset: 1675712280
```
This is the legacy format, and the reset number is obviously not correct.
Now the response headers look like this:
```
RateLimit-Limit: 20
RateLimit-Remaining: 7
RateLimit-Reset: 21
```
Using the new format and correct reset value.
And if a request has been rate limited, includes a `Retry-After`:
```
RateLimit-Limit: 20
RateLimit-Remaining: 0
RateLimit-Reset: 18
Retry-After: 60
```

### DIFF
--- a/src/provisioning/api.ts
+++ b/src/provisioning/api.ts
@@ -145,8 +145,15 @@ export class ProvisioningApi {
         this.app.get('/health', this.getHealth.bind(this));
 
         const limiter = this.opts.ratelimit && ratelimiter({
-            handler: (req, _res, next) => {
-                next(new ApiError("Too many requests", ErrCode.Ratelimited, 429));
+            handler: (req, _res, next, options) => {
+                next(new ApiError(
+                    "Too many requests",
+                    ErrCode.Ratelimited,
+                    429,
+                    {
+                        retry_after_ms: options.windowMs,
+                    }
+                ));
             },
             windowMs: 1 * 60 * 1000, // 1 minute
             max: 30, // Limit per window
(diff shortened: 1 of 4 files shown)


src/provisioning/api.ts
```diff
@@ -146,13 +146,13 @@ export class ProvisioningApi {
 
         const limiter = this.opts.ratelimit && ratelimiter({
             handler: (req, _res, next) => {
-                const info = (req as AugmentedRequest).ratelimit as RateLimitInfo;
-                const retryAfterMs = info?.resetTime ? info.resetTime.getTime() - Date.now() : null;
-                next(new ApiError("Too many requests", ErrCode.Ratelimited, 429, { retry_after_ms: retryAfterMs }));
```
Half-Shot, line 151: This *is* a departure from what we do in the Matrix spec, although I am happy to go along with what the HTTP standard is. We aren't beholden to Matrix here.

Half-Shot (approved):

src/provisioning/api.ts
justinbot: Fixed so it has both :)

PR #459, opened 2023-02-28
Migrate the ban-sync component from matrix-appservice-irc

Half-Shot:
https://github.com/matrix-org/matrix-appservice-irc/blob/develop/src/bridge/MatrixBanSync.ts never really needed to live in the IRC bridge and in practice might be useful for other bridges.

I've been cheeky here and added an additional feature to check for open registration in the code.

The tests still need porting.

### DIFF
--- a/src/components/ban-sync.ts
+++ b/src/components/ban-sync.ts
@@ -1,0 +1,143 @@
+
+
+import { Intent, MatrixUser, WeakStateEvent, Logger } from "..";
+import { MatrixGlob } from "matrix-bot-sdk";
+
+const log = new Logger("MatrixBanSync");
+
+export interface MatrixBanSyncConfig {
+    rooms: string[];
+}
+
+enum BanEntityType {
+    Server = "m.policy.rule.server",
+    User = "m.policy.rule.user"
+}
+
+interface BanEntity {
+    matcher: MatrixGlob;
+    entityType: BanEntityType;
+    reason: string;
+}
+
+interface MPolicyContent {
+    entity: string;
+    reason: string;
+    recommendation: "m.ban";
+}
+
+function eventTypeToBanEntityType(eventType: string): BanEntityType|null {
+    switch (eventType) {
+        case "m.policy.rule.user":
+        case "org.matrix.mjolnir.rule.user":
+            return BanEntityType.User;
+        case "m.policy.rule.server":
+        case "org.matrix.mjolnir.rule.server":
+            return BanEntityType.Server
+        default:
+            return null;
+    }
+}
+
+const supportedRecommendations = [
+    "org.matrix.mjolnir.ban", // Used historically.
+    "m.ban"
+];
+/**
+ * Synchronises Matrix `m.policy.rule` events with the bridge to filter specific
+ * users from using the service.
+ */
+export class MatrixBanSync {
+    private bannedEntites = new Map<string, BanEntity>();
+    private subscribedRooms = new Set<string>();
+    constructor(private config: MatrixBanSyncConfig) { }
+
+    public async syncRules(intent: Intent) {
+        this.bannedEntites.clear();
+        this.subscribedRooms.clear();
+        for (const roomIdOrAlias of this.config.rooms) {
+            try {
+                const roomId = await intent.join(roomIdOrAlias);
+                this.subscribedRooms.add(roomId);
+                const roomState = await intent.roomState(roomId, false) as WeakStateEvent[];
+                for (const evt of roomState) {
+                    this.handleIncomingState(evt, roomId);
+                }
+            }
+            catch (ex) {
+                log.error(`Failed to read ban list from ${roomIdOrAlias}`, ex);
+            }
+        }
+    }
+
+    /**
+     * Is the given room considered part of the bridge's ban list set.
+     * @param roomId A Matrix room ID.
+     * @returns true if state should be handled from the room, false otherwise.
+     */
+    public isTrackingRoomState(roomId: string): boolean {
+        return this.subscribedRooms.has(roomId);
+    }
+
+    /**
+     * Checks to see if the incoming state is a recommendation entry.
+     * @param evt A Matrix state event. Unknown state events will be filtered out.
+     * @param roomId The Matrix roomID where the event came from.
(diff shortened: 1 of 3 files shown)


src/components/ban-sync.ts
```diff
@@ -0,0 +1,249 @@
...
+export class MatrixBanSync {
+    private readonly homeserverPropertiesCache = new Map<string, {openRegistration: RegistrationStatus; ts: number;}>();
+    private readonly bannedEntites = new Map<string, BanEntity>();
+    private readonly subscribedRooms = new Set<string>();
+    private readonly hostResolver = new MatrixHostResolver();
+    constructor(private config: MatrixBanSyncConfig) {
+
+    }
+
+    public async getHomeserverProperties(serverName: string) {
```
justinbot: Could use jsdoc and return type.

src/components/ban-sync.ts
```diff
@@ -0,0 +1,249 @@
...
+        const hsData = this.homeserverPropertiesCache.get(serverName);
+        // Slightly fuzz the ttl.
+        const ttl = CACHE_HOMESERVER_PROPERTIES_FOR_MS + (Math.random()*60000);
```
justinbot: I think this is missing a `Date.now()`?

src/components/ban-sync.ts
```diff
@@ -0,0 +1,249 @@
...
+        if (hsData && hsData.ts < ttl) {
+            return hsData;
+        }
+
+        const { url } = await this.hostResolver.resolveMatrixServer(serverName);
+        const registrationResponse = await axios.post(new URL('/_matrix/client/v3/register', url).toString(), { }, { });
+
+        let openReg = RegistrationStatus.Unknown;
```
justinbot: I think it would help readability to put this all this logic in a separate method which only handles returning a `RegistrationStatus`. Also that way it could `return` instead of mutating `openReg`.

src/components/ban-sync.ts
```diff
@@ -0,0 +1,249 @@
...
+
+        const hsProps = {
+            openRegistration: openReg,
+            ts: Date.now(),
+        };
+        this.homeserverPropertiesCache.set(serverName, hsProps);
+        return hsProps;
+    }
+
+    public async syncRules(intent: Intent) {
```
justinbot, line 176: Could use a jsdoc and return type.

src/components/ban-sync.ts
```diff
@@ -0,0 +1,249 @@
...
+                // Invalid response
+                openReg = RegistrationStatus.Unknown;
+            }
+            else if (flows.length === 0) {
+                // No available flows, so closed.
+                openReg = RegistrationStatus.Closed;
+            }
+            else {
+                // Check the flows
+                for (const flow of flows) {
```
justinbot: Finding this loop a bit hard to follow, I think it could be more straightforward with a `.map` of flow stages to `RegistrationStatus` and then taking the minimum?

*Just a suggestion feel free to ignore :)*

justinbot (approved):
Some readability suggestions but overall looks good.

src/components/ban-sync.ts
```diff
@@ -43,19 +49,106 @@ const supportedRecommendations = [
...
+    public async getHomeserverProperties(serverName: string) {
+        const hsData = this.homeserverPropertiesCache.get(serverName);
+        // Slightly fuzz the ttl.
+        const ttl = CACHE_HOMESERVER_PROPERTIES_FOR_MS + (Math.random()*60000);
+        if (hsData && hsData.ts < ttl) {
+            return hsData;
+        }
+
+        const { url } = await this.hostResolver.resolveMatrixServer(serverName);
+        const registrationResponse = await axios.post(new URL('/_matrix/client/v3/register', url).toString(), { }, { });
```
thibaultamartin: What about older homeservers? If old homeservers only expose `/_matrix/client/r0/register`, will the bridge allow them to stay despite having open reg?

src/components/ban-sync.ts
```diff
@@ -43,19 +49,106 @@ const supportedRecommendations = [
...
+            // Look at the flows.
+            const { flows } = registrationResponse.data as MatrixRegistrationResponse;
+            if (!flows) {
+                // Invalid response
+                openReg = RegistrationStatus.Unknown;
+            }
+            else if (flows.length === 0) {
+                // No available flows, so closed.
+                openReg = RegistrationStatus.Closed;
+            }
```
thibaultamartin: In the case of GNOME, I tried

```
>>> r = requests.post("https://gnome.modular.im/_matrix/client/v3/register", json={})
>>> r.json()
{'session': 'lpztxwMUAyTzgzdLHWhvwGtm', 'flows': [{'stages': ['m.login.recaptcha', 'm.login.terms', 'm.login.email.identity']}], 'params': {'m.login.recaptcha': {'public_key': '6LcgI54UAAAAABGdGmruw6DdOocFpYVdjYBRe4zb'}, 'm.login.terms': {'policies': {'privacy_policy': {'version': '1.0', 'en': {'name': 'Privacy Policy', 'url': 'https://gnome.modular.im/_matrix/consent?v=1.0'}}}}}}
>>> r.json()["flows"]
[{'stages': ['m.login.recaptcha', 'm.login.terms', 'm.login.email.identity']}]
```

It looks like the HS has open registrations, but in practice it's restricted to `xxx@gnome.org` email addresses. Is there a way to detect that remotely?

src/components/ban-sync.ts
Half-Shot: Fair point, we'll include those.

src/components/ban-sync.ts
Half-Shot: Added.

src/components/ban-sync.ts
Half-Shot: Done.

src/components/ban-sync.ts
Half-Shot: Done.

Half-Shot:
Hey there people on the tail end of 2025, closing this PR as we've in practice not felt the need to work on it over the last few years. I suspect this feature is still quite useful, however my focus on the IRC bridge has tailed off due to demands elsewhere and the foundation would need to pick this one up.

PR #460, opened 2023-03-05
Add return type to createInitialState

progval:


progval:
Signed-off-by: Valentin Lorentz \<progval plus git at progval dot net\>

src/components/bridge-info-state.ts
```diff
@@ -35,6 +35,12 @@ export interface MSC2346Content extends MappingInfo {
     bridgebot: string;
 }
 
+export interface InitialEvent {
+    type: string,
+    content: Record<string, unknown>,
```
Half-Shot, line 40: Isn't this just `MSC2346Content`?

Half-Shot (commented):
Confusion on types

src/components/bridge-info-state.ts
progval: Ah yes sorry, I should have explained that. First, because callers need to cast it to a record of unknown anyway: https://github.com/matrix-org/matrix-appservice-irc/blob/3158cc521140ca6883e157ddb0bf01e96d455af4/src/provisioning/Provisioner.ts#L582

and also because I am working on adding support for MSC3968 and MSC3969, which will need genericity in the record type

PR #462, opened 2023-03-07
Fix provisioning token parsing

justinbot:
Removes lowercasing the header which would break auth when the provisioning token contains uppercase characters.

### DIFF
--- a/src/provisioning/api.ts
+++ b/src/provisioning/api.ts
@@ -243,11 +243,11 @@ export class ProvisioningApi {
         // Historically, user_id has been used. The bridge library supports either.
         // eslint-disable-next-line camelcase
         req: Request<unknown, unknown, {userId?: string, user_id?: string}>, res: Response, next: NextFunction) {
-        const authHeader = req.header("Authorization")?.toLowerCase();
+        const authHeader = req.header("Authorization");
         if (!authHeader) {
             throw new ApiError('No Authorization header', ErrCode.BadToken);
         }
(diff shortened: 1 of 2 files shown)


src/provisioning/api.ts
```diff
@@ -243,11 +243,11 @@ export class ProvisioningApi {
         // Historically, user_id has been used. The bridge library supports either.
         // eslint-disable-next-line camelcase
         req: Request<unknown, unknown, {userId?: string, user_id?: string}>, res: Response, next: NextFunction) {
-        const authHeader = req.header("Authorization")?.toLowerCase();
+        const authHeader = req.header("Authorization");
         if (!authHeader) {
             throw new ApiError('No Authorization header', ErrCode.BadToken);
         }
-        const token = authHeader.startsWith("bearer ") && authHeader.substring("bearer ".length);
+        const token = authHeader.replace("Bearer ", "").replace("bearer ", "");
```
Half-Shot, line 250: Of course, you might write bEaReR (I know, massive edge case, but still) and this wouldn't work. Perhaps we can just do `authHeader.toLowerCase().startsWith("bearer ")`?

justinbot: Oauth RFC says it must be `Bearer` with uppercase `B`: https://www.rfc-editor.org/rfc/rfc6750#section-2.1

Half-Shot: Oh, in that case, screw people that do anything else :tada:

Half-Shot (approved):

PR #463, opened 2023-03-12
Add support for MSC3968 ('poorer features')

progval:
This is a breaking change, as it makes `createInitialState` return a
list of events instead of a single event.

No new test, because components don't have any.

https://github.com/matrix-org/matrix-appservice-irc/pull/1686 uses it and is tested for PMs

### DIFF
--- a/src/components/bridge-info-state.ts
+++ b/src/components/bridge-info-state.ts
@@ -1,5 +1,6 @@
 import { Bridge } from "../bridge";
 import { Logger } from "..";
+import { Intent } from "./intent";
 import PQueue from "p-queue";
 
 const log = new Logger("BridgeStateSyncer");
@@ -29,11 +30,30 @@ export interface MappingInfo {
         // eslint-disable-next-line camelcase
         external_url?: string;
     };
+    eventFeatures?: MSC3968Content;
 }
 
 export interface MSC2346Content extends MappingInfo {
     bridgebot: string;
 }
+
+export interface MSC3968Content {
+    keys_default?: number,
+    keys?: Record<string, number>,
+    html_elements_default?: number,
+    html_elements?: Record<string, number>
+    mimetypes_default?: number,
+    mimetypes?: Record<string, number>
+    msgtypes_default?: number,
+    msgtypes?: Record<string, number>
+}
+
+export interface InitialEvent {
+    type: string,
+    content: Record<string, unknown>,
+    state_key: string,
+}
+
 
 interface Opts<BridgeMappingInfo> {
     /**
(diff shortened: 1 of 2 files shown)


progval:
Signed-off-by: Valentin Lorentz \<progval plus matrixsignoff at progval dot net\>

src/components/bridge-info-state.ts
```diff
@@ -140,4 +206,12 @@ export class BridgeInfoStateSyncer<BridgeMappingInfo> {
...
         return content;
     }
+
+    public createEventFeaturesContent(mapping: MappingInfo)
+    : MSC3968Content|null {
+        if (!mapping.eventFeatures) {
+            return null;
+        }
+        return mapping.eventFeatures;
+    }
```
Half-Shot: Create implies this isn't atomic, when really we're just returning some info. May I offer you a getter.

```suggestion
    public getEventFeaturesContent(mapping: MappingInfo)
    : MSC3968Content|null {
        return mapping?.eventFeatures ?? null;
    }
```

progval: Aren't getters supposed to take no parameters?

Half-Shot: It would be so helpful if I could read. Have removed a space.

src/components/bridge-info-state.ts
```diff
@@ -78,45 +99,90 @@ export class BridgeInfoStateSyncer<BridgeMappingInfo> {
...
-                    roomId, BridgeInfoStateSyncer.EventType, key, true
-                );
-                if (eventData !== null) { // If found, validate.
-                    if (JSON.stringify(eventData) === JSON.stringify(content)) {
-                        continue;
-                    }
-                    log.debug(`${key} for ${roomId} is invalid, updating`);
-                }
+            const bridgeInfoContent = this.createBridgeInfoContent(realMapping);
+            const eventFeaturesContent = this.createEventFeaturesContent(realMapping);
```
progval:
```suggestion
            const eventFeaturesContent = this.getEventFeaturesContent(realMapping);
```

src/components/bridge-info-state.ts
```diff
@@ -78,45 +99,90 @@ export class BridgeInfoStateSyncer<BridgeMappingInfo> {
...
-        };
+        const events = [
+            {
+                type: BridgeInfoStateSyncer.EventType,
+                content: this.createBridgeInfoContent(mapping) as unknown as Record<string, unknown>,
+                state_key: this.createStateKey(mapping),
+            },
+        ];
+
+        const eventFeaturesContent = this.createEventFeaturesContent(mapping)
```
progval:
```suggestion
        const eventFeaturesContent = this.getEventFeaturesContent(mapping)
```

src/components/bridge-info-state.ts
progval: Sorry for the delay applying this, I didn't understand what you meant by "Have removed a space." and didn't notice the edit

progval:
ping?

PR #465, opened 2023-03-27
Ensure every route is caught by onError when using the provisioning API

Half-Shot:

### DIFF
--- a/src/provisioning/api.ts
+++ b/src/provisioning/api.ts
@@ -220,7 +220,7 @@ export class ProvisioningApi {
         path: string,
         handler: (req: ProvisioningRequest, res: Response, next?: NextFunction) => void|Promise<void>,
         fnName?: string): void {
-        this.baseRoute[method](path, async (req, res, next) => {
+        this.baseRoute[method](path, async (req: Express.Request, res: Response, next: NextFunction) => {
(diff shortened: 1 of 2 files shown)


src/provisioning/api.ts
```diff
@@ -236,7 +236,8 @@ export class ProvisioningApi {
                 // Pass to error handler.
                 next([ex, provisioningRequest]);
             }
-        });
+            // Always add an error handler
+        }, this.onError);
```
justinbot, line 240: ~~Just noting this will supersede any error handling implementation in the sublcass~~
e.g. https://github.com/matrix-org/matrix-appservice-irc/blob/develop/src/provisioning/Provisioner.ts#L144-L158

(Actually looking again I suppose it wouldn't since that wrapper goes before this error handler?)

I suppose the idea is you would now override `onError` instead?

justinbot (approved):
Looks good, just a note on how this affects existing usages.

src/provisioning/api.ts
Half-Shot: Yeah, that's certainly one way. Or you could wrap your handler in a try/catch and just not emit an error. The onError code was always intended to catch errors that fell through routes.

PR #466, opened 2023-04-19
Add support for Node 20, and drop support for Node 16

Half-Shot:
As per https://github.com/nodejs/release#release-schedule, Node 20 has now been released and we will start testing for it. Node 16 is in maintenance and will no longer be supported. Node 18 is the last version we will support.


package.json
```diff
@@ -53,17 +52,19 @@
...
     "@types/jasmine": "^4.0.3",
     "@types/js-yaml": "^4.0.0",
     "@types/nedb": "^1.8.11",
-    "@types/node": "^16",
+    "@types/node": "^18",
     "@types/nopt": "^3.0.29",
     "@types/uuid": "^8.3.1",
     "@typescript-eslint/eslint-plugin": "^5.25.0",
     "@typescript-eslint/parser": "^5.25.0",
+    "@tsconfig/node18": "^1",
```
justinbot: Looks like this is on `2.0.0`, was there an issue with that version? https://www.npmjs.com/package/@tsconfig/node18

justinbot (approved):
One comment but looks good (looks like CI will be fixed by upgrading `matrix-appservice` after https://github.com/matrix-org/matrix-appservice-node/pull/65)

package.json
Half-Shot: Ah, I just mindlessly bumped. Fix'd.

AndrewFerr:
Is now a good time to replace the `npm` commands in `package.json` (and docs files) with their `yarn` equivalents?

package.json
```diff
@@ -70,7 +71,7 @@
     "nyc": "^15.1.0",
     "ts-node": "^10.2.1",
     "typedoc": "^0.23.7",
-    "typescript": "^4.4.3",
+    "typescript": "^5.0.4",
```
AndrewFerr, line 74: This creates a peer dependency failure when installing `typedoc`.

Fortunately, [`typedoc` supports Typescript 5 as of version `0.23.28`](https://github.com/TypeStrong/typedoc/issues/2201), so we can either upgrade to that version or the latest one (currently `0.24.6`).  We should probably go with the former since the latter is bleeding-edge (released 13 hours ago!).

I already have a commit that makes this update, and can push it here if that's fine.

AndrewFerr:
The dependencies of the two example modules (under [./examples](https://github.com/matrix-org/matrix-appservice-bridge/tree/develop/examples)) may as well be updated too.  It turns out their lockfiles were quite out of whack, too.  If it's welcomed, I prepared a commit to fix those issues that could be pushed here.

Half-Shot:
>The dependencies of the two example modules (under [./examples](https://github.com/matrix-org/matrix-appservice-bridge/tree/develop/examples)) may as well be updated too. It turns out their lockfiles were quite out of whack, too. If it's welcomed, I prepared a commit to fix those issues that could be pushed here.

Feel free, though longer term we need to sort out those examples. I think we always forget about them and CI don't care :p.

.github/workflows/test.yml
```diff
@@ -7,20 +7,16 @@ on:
 jobs:
   lint:
     runs-on: ubuntu-20.04
```
AndrewFerr, line 9: Is there a reason the test workflow runs on `20.04` but the release & publish workflows run on `latest`?

Half-Shot: The tests need to run on LTS-1 because that's last supported, the workflows that don't test / build native extensions run on latest because it's the default, but they could easily run on 20.04 for safety.

package.json
```diff
@@ -53,14 +52,16 @@
...
   "devDependencies": {
+    "@tsconfig/node18": "^2.0.0",
     "@types/cors": "^2.8.12",
     "@types/express": "^4.17.14",
     "@types/extend": "^3.0.1",
     "@types/jasmine": "^4.0.3",
     "@types/js-yaml": "^4.0.0",
     "@types/nedb": "^1.8.11",
-    "@types/node": "^16",
+    "@types/node": "^18",
```
AndrewFerr, line 62: Likewise, `package.json` should have its `"engines"` section updated to `"node": ">=18"` like it was for [matrix-appservice-node](https://github.com/matrix-org/matrix-appservice-node/commit/3413960453c63f16e79de95c1979ff5d25b71d6f#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R8).

AndrewFerr (approved):
With the glibc compat issues resolved, this is finally done :smiley:

PR #468, opened 2023-05-02
Fix: leave should not execute kick as user can't kick itself.

iwalkalone69:
When a user leaves an IRC channel this should execute leaveRoom instead of kickUser even if it has a reason. Otherwise it fails because user has no permission to kick himself.

### DIFF
--- a/src/components/intent.ts
+++ b/src/components/intent.ts
@@ -586,7 +586,7 @@ export class Intent {
      */
     public async leave(roomId: string, reason?: string) {
         await this.botSdkIntent.ensureRegistered();
-        return this.botSdkIntent.leaveRoom(roomId);
+        return this.botSdkIntent.leaveRoom(roomId, reason);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


Half-Shot:
We introduced this as a way to leave with a reason, because it at least used to be the case that you can always kick yourself (which at a high level is the same as leaving, just with the option to leave a departing message). These days the spec supports leaving with a reason, so we should allow that.

src/components/intent.ts
```diff
@@ -585,10 +585,7 @@ export class Intent {
      * @param reason An optional string to explain why the user left the room.
      */
     public async leave(roomId: string, reason?: string) {
-        if (reason) {
-            await this.botSdkIntent.ensureRegistered();
-            return this.botSdkIntent.underlyingClient.kickUser(this.userId, roomId, reason);
-        }
+        await this.botSdkIntent.ensureRegistered();
         return this.botSdkIntent.leaveRoom(roomId);
```
Half-Shot:
```suggestion
        return this.botSdkIntent.leaveRoom(roomId, reason);
```

Half-Shot (approved):
Otherwise LGTM.

**Edit**: And please sign off / add a changelog :)

PR #469, opened 2023-05-02
Run tests on PRs too

Half-Shot:
Ensures PRs on forked branches have the tests run too.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,7 +6,6 @@ on:
       - develop
   workflow_dispatch:
   pull_request:
-    types: [opened, edited, synchronize]
 
 jobs:
   lint:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -2,7 +2,11 @@ name: Tests
 
 on:
   push:
+    branches:
+      - develop
   workflow_dispatch:
+  pull_request:
+    types: [opened, edited, synchronize]
```
justinbot: This can be left out to run on all PRs, like we have done elsewhere:
https://github.com/matrix-org/matrix-appservice-slack/blob/7a09bda9aa780bd3d9d01e091811e3356fb1e115/.github/workflows/tests.yml#L6

justinbot (approved):

PR #471, opened 2023-05-10
activity-tracker: Fix parsing of user domains

progval:
Server names can contain a colon: https://spec.matrix.org/v1.6/appendices/#server-name

### DIFF
--- a/src/components/activity-tracker.ts
+++ b/src/components/activity-tracker.ts
@@ -107,7 +107,7 @@ export class ActivityTracker {
             // Failed to get presence, going to fallback to admin api.
         }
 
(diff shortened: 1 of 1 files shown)


progval:
Signed-off-by: Valentin Lorentz \<progval plus matrixsignoff at progval dot net\>

Half-Shot (changes requested):
Seems okay, but needs a changelog. Ideally https://github.com/matrix-org/matrix-appservice-bridge/blob/11c447ed491a556b8e8d40e0b0a9fc7c90f71ccf/spec/unit/activity-tracker.spec.ts would be updated with a test case to check that these userids are still supported.

PR #498, opened 2024-05-24
Binary-encode media proxy URLs rather than base64'ing a JSON

tadzik:
Saves us 80 bytes for the test data (down from 208 to 128), while leaving room for future changes.


Half-Shot (approved):
Looks smart to me, no concerns!

PR #501, opened 2024-06-30
fix: spec compliant token verification

debdutdeb:

### DIFF
--- a/src/bridge.ts
+++ b/src/bridge.ts
@@ -1582,13 +1582,23 @@ export class Bridge {
             // Bridge isn't ready yet
             return false;
         }
-        if (
-            req.query.access_token !== this.registration.getHomeserverToken() &&
-            req.get("authorization") !== `Bearer ${this.registration.getHomeserverToken()}`
-        ) {
-            return false;
+
+        const tokenFromQuery = req.query.access_token;
+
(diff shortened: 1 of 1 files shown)


src/bridge.ts
```diff
@@ -1740,3 +1750,5 @@ async function loadDatabase<T extends BridgeStore>(path: string, Cls: new (db: D
         throw Error('nedb could not be imported. You will need to add this package as a peer dependency.');
     }
 }
+
+// vi: et sw=4 ts=4
```
tadzik, line 1754: Looks like an accidental leftover

tadzik (commented):
LGTM aside of the `vi` marker

src/bridge.ts
```diff
@@ -1582,13 +1582,23 @@ export class Bridge {
...
         }
-        if (
-            req.query.access_token !== this.registration.getHomeserverToken() &&
-            req.get("authorization") !== `Bearer ${this.registration.getHomeserverToken()}`
-        ) {
-            return false;
+
+        const tokenFromQuery = req.query.access_token;
+
+        const tokenFromHeader = req.get("authorization").substring(7); // "Bearer ".length === 7
```
Half-Shot, line 1588: Should we check for "Bearer "?

PR #502, opened 2024-07-24
update CONTRIBUTING.md with DCO

bluesomewhere:
The Matrix.org Foundation is rolling out this DCO across all public repositories in its namespace. Of note, this DCO does not require the use of "real" or "legally identifiable" names.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -91,8 +91,7 @@ When making a pull request, please ensure it [the PR] follows these best practis
 
 - Targets `develop` (unless it explicitly depends upon another feature, then depend on that branch and comment to that effect in the PR body).
 - Is updated via rebase mechanisms when `develop` changes, rather than merge commits (reduces noise).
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -125,6 +125,59 @@ When reviewing a PR, a maintainer will:
    they have acknowledged the message and will make the change.
  - Approve a PR which is ready to merge, or nearly ready with some minor tweaks or optional improvements.
 
+### Sign off
```
Half-Shot, line 127: N.B. just to mention that we link to the Synapse sign off page above. You might wanna adjust that link. https://github.com/matrix-org/matrix-appservice-bridge/pull/502/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055R94

Half-Shot (approved):
Good modulo the link

CONTRIBUTING.md
bluesomewhere: Oh yes good catch, this is as good a time as any to update that link. Thank you!

Issue #433, opened 2022-08-07
HOWTO.md: Replace "node:querystring" with URLSearchParams API


jaller94:
The querystring API is considered Legacy. new code should use the URLSearchParams API instead.

https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams

This is fully supported since Node.js 10.

PR #503, opened 2024-08-09
Update HOWTO.md

AllMightLegend:
#433 Issue
Replaced "node:querystring" with URLSearchParams API


tadzik (changes requested):
This does not look correct. You're passing the `http`  module as a query string to parse, and then using a boolean method to parse.

This should look more like `const params= new URLSearchParams(body)`, and then `params.get('user_name')` etc.

PR #504, opened 2024-08-14
Fix MediaProxy to make it actually start up

tadzik:
Slipped through the cracks in https://github.com/matrix-org/matrix-appservice-bridge/pull/444, came up when testing the IRC integration (https://github.com/matrix-org/matrix-appservice-irc/pull/1805)


Half-Shot (approved):
oopsie whoopsie

tadzik:
Deliberately skipping the changelog, since we're just fixing up the other PR.

PR #506, opened 2024-09-05
Use the new authenticated endpoints in MediaProxy

tadzik:
Fixes the requests failing in https://github.com/matrix-org/matrix-appservice-irc/issues/1816

### DIFF
--- a/src/components/media-proxy.ts
+++ b/src/components/media-proxy.ts
@@ -147,7 +147,12 @@ export class MediaProxy {
         }
         // Cache from this point onwards.
         // Extract the media from the event.
-        const url = this.matrixClient.mxcToHttp('mxc://' + metadata.mxc);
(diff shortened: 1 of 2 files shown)


src/components/media-proxy.ts
```diff
@@ -147,7 +147,12 @@ export class MediaProxy {
         }
         // Cache from this point onwards.
         // Extract the media from the event.
-        const url = this.matrixClient.mxcToHttp('mxc://' + metadata.mxc);
+        const mxcMatch = metadata.mxc.match(new RegExp('^([^/]+)/(.+)$'));
```
Half-Shot, line 150: Assume we have tests that cover this :)

Half-Shot (approved):

src/components/media-proxy.ts
tadzik: Manual ones, I'm afraid – our tests never get deep enough to hit onMediaRequest.

I can pull this out into a module and test it there.

PR #521, opened 2025-09-02
Fix nedb's usage of `node:utils` for Node 24

AndrewFerr:
Some parts of `nedb` don't work in Node 24 due to calling removed `node:utils` APIs.

https://github.com/matrix-org/matrix-appservice-bridge/pull/519 bypassed that problem in unit tests by injecting the required `node:utils` functions, but that doesn't fix the problem in production or for anything that uses this SDK as a dependency.

This PR moves that injection into source code to fix the problem outside of just tests.  It also adds a deprecation warning log to make it clear that NeDB is no longer supported & should not be used anymore.

### DIFF
--- a/spec/integ/bridge.spec.js
+++ b/spec/integ/bridge.spec.js
@@ -2,11 +2,6 @@
 const Datastore = require("nedb");
 const fs = require("fs");
 
-// required fix for nedb being incredibly outdated
-const util = require("node:util");
-util.isDate = util.types.isDate;
-util.isRegExp = util.types.isRegExp;
-
 const HS_URL = "http://example.com";
 const HS_DOMAIN = "example.com";
 const BOT_LOCALPART = "the_bridge";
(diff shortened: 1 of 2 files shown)


Half-Shot (commented):
Gooooooood god what a hack.

I'd be minded at this stage to rip out NeDB support entirely? How much work would that be?

AndrewFerr:
> I'd be minded at this stage to rip out NeDB support entirely? How much work would that be?

It probably won't be much trouble, if at all -- the only concern is whether there should be a tool/script to migrate from NeDB to something else.

Half-Shot:
I think we built those out for a few bridges, and we've supported postgres for ages. I think it might be better to remove it at this point, it's super old.

AndrewFerr:
Not needed anymore thanks to https://github.com/matrix-org/matrix-appservice-bridge/pull/528 :tada:

PR #528, opened 2026-07-30
Remove NeDB

### LINKED ISSUE #77: Nedb is unmaintained, alternatives needed.

Half-Shot:
Fixes https://github.com/matrix-org/matrix-appservice-bridge/issues/77

This removes NeDB entirely from the bridge and fixes a long standing security bug. Consumers of the bridge will now need to provide their own Datastore-like if they are going to use the storage classes.


AndrewFerr (approved):
Keeping an equivalent interface that consumers must implement is a good idea.

AndrewFerr (approved):

